PR #5, opened 2022-02-21
Solution

Geksus:


app/managers.py
```diff
@@ -0,0 +1,43 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self):
+        self._connection = sqlite3.connect("cinema.db3")
+        self.table_name = "actor"
```
Nazar-kob, line 9: It would be better to create a table called "actors" instead of "actor", but under the conditions of the task everything is correct, so I approve!

Nazar-kob (approved):

MasterpieceElbow (approved):

PR #19, opened 2022-02-22
'Solution'

Oleksandr563487953:


Geksus (changes requested):
You did not commit managers.py, so it's impossible to tell wether your solution is correct. Try adding managers.py to your next commit

PR #28, opened 2022-06-15
Solution

YuriiKindrat:


app/managers.py
```diff
@@ -0,0 +1,45 @@
...
+class ActorManager:
+
+    def __init__(self):
+        self.connection = sqlite3.connect("cinema_db.db3")
+        self.name_of_db = "actors"
+
+    def all(self):
+
+        cinema_data_cursor = self.connection.execute(
+            f"SELECT id, first_name, last_name FROM {self.name_of_db}"
```
CheshkoSergey, line 15: it is better to use SELECT * instead SELECT id, first_name, last_name

CheshkoSergey (approved):

PR #29, opened 2022-06-15
Solution

Magus6784:


Hectorovich (approved):

app/managers.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def all(self):
+        actor_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name FROM {self.table_name}"
+        )
+
+        return [Actor(*row) for row in actor_cursor]
+
+    def update(self, id_to_update: int,
+               new_first_name_: str, new_last_name: str):
```
TPolina, line 26: What for do you add `_` at the end of a variable name?

TPolina (approved):

app/managers.py
Magus6784: Samurai have no goal, only way

PR #35, opened 2022-06-15
Solution

Hectorovich:


Flashmobber (approved):

app/managers.py
```diff
@@ -0,0 +1,37 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self):
+        self._connection = sqlite3.connect("cinema.db3")
+        self.table_name = "cinemas"
```
TPolina, line 9: why do you have a table called `cinemas` if you store information about the actors in it?

app/managers.py
```diff
@@ -0,0 +1,37 @@
...
+        )
+        self._connection.commit()
+
+    def all(self):
+        cinemas_cursor = self._connection.execute(
+            f"SELECT * FROM {self.table_name}"
+        )
+        return [Actor(*row) for row in cinemas_cursor]
+
+    def update(self, new_id: int, new_first_name: str, new_last_name: str):
```
TPolina, line 24: you call the parameter `new_id`, although in fact id is old and it does not change

TPolina (changes requested):

PR #37, opened 2022-06-15
Solution

Chardona:


app/managers.py
```diff
@@ -0,0 +1,39 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self):
+        self._con = sqlite3.connect("cinema.db")
```
KatiukhaO, line 8: i think, that, you could better named variable _con, more readable

in everything else, nice code

KatiukhaO (approved):

app/managers.py
Chardona: thanks.

app/managers.py
```diff
@@ -0,0 +1,39 @@
...
+        )
+
+        return [Actor(*row) for row in actor_cursor]
+
+    def update(self, id_: int, new_name: str, new_surname: str):
+        self._con.execute(
+            f"UPDATE {self.table_name}"
+            "SET first_name = ?, last_name = ?"
+            "WHERE id = ?",
+            (id_, new_name, new_surname)
```
TPolina, line 31: Incorrect order of parameters, so `update` doesn't work correctly.

app/managers.py
```diff
@@ -0,0 +1,39 @@
...
+        )
+        self._con.commit()
+
+    def delete(self, id_: int):
+        self._con.execute(
+            f"Delete from {self.table_name} WHERE id = ?", (id_,)
```
TPolina, line 37: use uppercase in `delete from` as well

TPolina (changes requested):

TPolina (approved):

PR #48, opened 2022-06-16
'Solution'

daria-niklaevna:


app/managers.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class ActorManager:
+    def __init__(self):
+        self._connection = sqlite3.connect("cinema.sqlite")
+        self.table_name = "actor"
+
+    def all(self):
+        actor_manager_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name FROM {self.table_name}"
```
katevasilenko, line 12: you can write * instead (but actually I wrote this way to))

katevasilenko (approved):

TPolina (approved):

PR #51, opened 2022-06-16
Solution

BohdanKryven:


app/managers.py
```diff
@@ -0,0 +1,40 @@
...
+            f"SET first_name = ?, last_name = ?"
+            f"WHERE id = ?",
+            (first_name, last_name, which_id_to_update)
+        )
+        self._connection.commit()
+
+    def delete(self, which_id_to_delete):
+        self._connection.execute(
+            f"DELETE FROM {self._table_name} "
+            f"WHERE id = {which_id_to_delete}"
```
TPolina, line 38: use parametrized query here as well

TPolina (changes requested):

app/managers.py
BohdanKryven: Thanks. Fixed

TPolina (approved):

PR #52, opened 2022-06-16
Develop

Kiawami:


app/managers.py
```diff
@@ -0,0 +1,43 @@
...
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self):
+        self._connection = sqlite3.connect("cinema.sqlite")
+
+    def create(self, first_name: str, last_name: str):
+        self._connection.execute(
+            "INSERT INTO actor (first_name, last_name) "
```
TPolina, line 12: Imagine you need to change table name. It would be a lot of work. Create variable `table_name` in `__init__` and use it.

TPolina (changes requested):

TPolina (approved):

PR #53, opened 2022-06-16
Solution

Evgeniy994:


app/managers.py
```diff
@@ -0,0 +1,43 @@
...
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self):
+        self._connection = sqlite3.connect("cinema.db3")
+        self.table_name = "actor"
+
+    def create(self, first_name: str, last_name: str):
+        self._connection.execute(
```
darten360, line 12: It's a bad style to use first paranthese near word and second one on a new line

app/managers.py
```diff
@@ -0,0 +1,43 @@
...
+            f"INSERT INTO {self.table_name} "
+            "(first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self):
+        actor_cursor = self._connection.execute(
```
darten360, line 20: the same

app/managers.py
```diff
@@ -0,0 +1,43 @@
...
+            f"SELECT * FROM {self.table_name}"
+        )
+
+        return [Actor(*row) for row in actor_cursor]
+
+    def update(
+            self, id_to_update: int, new_first_name: str, new_last_name: str
+    ):
+        self._connection.execute(
```
darten360, line 29: I think you understood me

darten360 (changes requested):

app/managers.py
Evgeniy994: looks like done)

Magus6784 (approved):

MasterpieceElbow (changes requested):
Remove unnecessary file `identifier.sqlite` from the pull request

PR #63, opened 2022-06-24
Solution

vsmutok:


app/models.py
```diff
@@ -0,0 +1,8 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class Actors:
```
IvanRamyk, line 5: The name of the class should be in the singular form

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #68, opened 2022-07-10
Solution

Splend1ed:


app/managers.py
```diff
@@ -0,0 +1,41 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self):
+        self.connection = sqlite3.connect("cinema.db")
+        self.table_name = "actor"
```
Abnormaltype, line 9: Does this DB contain only one author?

Abnormaltype (changes requested):

app/managers.py
Splend1ed: What u mean?

Abnormaltype: DB has the name `actor` which means, DB contains only one author

Abnormaltype (approved):
well done!

PR #69, opened 2022-07-11
Solution

CEHR2005:


app/managers.py
```diff
@@ -0,0 +1,42 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self):
+        self._connection = sqlite3.connect("cinema.db")
+        self.table_name = "actor"
```
Abnormaltype, line 9: Does this DB contain only one author?

Abnormaltype (changes requested):

PR #92, opened 2022-09-14
Solution

InnaKushnir:


app/managers.py
```diff
@@ -0,0 +1,39 @@
...
+class ActorManager:
+    def __init__(self):
+        self._connection = sqlite3.connect("cinema_db.db3")
+        self.table_name = "actors"
+
+    def all(self):
+        actors_cursor = self._connection.execute(
+            f"SELECT * FROM {self.table_name}"
+        )
+        print([Actor(*row) for row in actors_cursor])
```
Abnormaltype, line 14: do you need this?

Abnormaltype (changes requested):

app/managers.py
InnaKushnir: It was redundent

Abnormaltype (approved):
Well done!

PR #94, opened 2022-09-18
Solution

luxuriant777:


TPolina (approved):
Well done!

TPolina (changes requested):
Sorry, I didn't notice that you pushed db to GitHub. It's really important not to do so because it's not secure. Please, delete it.

PR #113, opened 2022-10-06
Develop

StasOkhrym:


app/managers.py
```diff
@@ -0,0 +1,50 @@
...
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.db3")
+        self.table_name = "actors"
+
+    def create(
```
mltdll, line 11: I don't think it's an issue, but wouldn't this method's definition fit in 79 characters?

app/managers.py
```diff
@@ -0,0 +1,50 @@
...
+        self._connection.commit()
+
+    def all(self) -> list[Actor]:
+        actors_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name "
+            f"FROM {self.table_name}"
+        )
+        return [Actor(*row) for row in actors_cursor]
+
+    def update(
```
mltdll, line 31: Same here.

mltdll (approved):

app/managers.py
```diff
@@ -0,0 +1,50 @@
...
+            (new_first_name, new_last_name, id_to_update)
+        )
+        self._connection.commit()
+
+    def delete(self, id_to_delete: int) -> None:
+        self._connection.execute(
+            f"DELETE FROM {self.table_name} "
+            f"WHERE id = ?",
+            (id_to_delete,)
+        )
```
Nattalli, line 50: I think it would be better to add `self._connection.commit()` here too

Nattalli (changes requested):

Nattalli (approved):
Well done!

PR #116, opened 2022-10-08
Solution for the task

olia-trofymchuk:


app/managers.py
```diff
@@ -0,0 +1,40 @@
...
+            f"VALUES (?, ?)",
+            (first_name, last_name,),
+        )
+        self._connection.commit()
+
+    def update(self, id_to_update: int,
+               first_name: str,
+               last_name: str) -> None:
+        self._connection.execute(
+            f"UPDATE {self.table_name} SET first_name = ?, last_name = ? "
```
mltdll, line 29: I'm not sure if SQL style guides should apply here, but SET clause is usually placed at new line.

app/managers.py
```diff
@@ -0,0 +1,40 @@
...
+            f"WHERE id = ?",
+            (first_name, last_name, id_to_update,)
+        )
+        self._connection.commit()
+
+    def delete(self, id_to_delete: int) -> None:
+        self._connection.execute(
+            f"DELETE FROM {self.table_name} WHERE id = ?",
```
mltdll, line 37: The same applies to WHERE.

mltdll (approved):

app/managers.py
Nattalli: yes, that is the best SQL practice. The same as using commands in the upper register! Good comment!

app/managers.py
```diff
@@ -0,0 +1,42 @@
...
+        cursor_with_actors = self._connection.execute(
+            f"SELECT * FROM {self.table_name}"
+        )
+        return [Actor(*row) for row in cursor_with_actors]
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} (first_name, last_name) "
+            f"VALUES (?, ?)",
+            (first_name, last_name,),
```
Nattalli, line 21: here is extra comma

app/managers.py
```diff
@@ -0,0 +1,42 @@
...
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.db3")
+        self.table_name = "actors"
+
+    def all(self) -> list:
```
Nattalli, line 11: `list[Actor]` would be better

app/managers.py
```diff
@@ -0,0 +1,42 @@
...
+        self._connection.commit()
+
+    def update(self, id_to_update: int,
+               first_name: str,
+               last_name: str) -> None:
+        self._connection.execute(
+            f"UPDATE {self.table_name} "
+            f"SET first_name = ?, last_name = ? "
+            f"WHERE id = ?",
+            (first_name, last_name, id_to_update,)
```
Nattalli, line 32: here extra comma too

Nattalli (changes requested):
Good job, but some comments left

app/managers.py
olia-trofymchuk: but is it not a tuple?

olia-trofymchuk: because in the video Dania used comma in the end of the tuple

Nattalli: I believe that for tuples next syntax is correct:
```python
empty_tuple = (,) 
one_item_in_tuple = (1,)
more_than_one = (1, 2) / (1, 2, 3)
```

olia-trofymchuk: Oh, i didn't know that :( 

Okay, thank you! i'll change my code)

Nattalli (approved):
Good job!

PR #118, opened 2022-10-09
Solution

IvanStored:


app/managers.py
```diff
@@ -0,0 +1,46 @@
...
+class ActorManager:
+
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.db3")
+        self.table_name = "actors"
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
+            f"(first_name, last_name) VALUES (?, ?)",
```
mltdll, line 16: It's not necessary to make this string an f-string. Same with lines 34, 35, 43.

mltdll (changes requested):

mltdll (approved):

MasterpieceElbow (approved):
GJ!

PR #121, opened 2022-10-13
Solution

kateryna3k:


app/managers.py
```diff
@@ -0,0 +1,38 @@
...
+            " VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def update(
+            self, id_to_update: int, new_first_name: str, new_last_name: str
+    ) -> None:
+        self._connection.execute(
+            f"UPDATE {self.table_name} SET first_name = ?, last_name = ?"
```
mltdll, line 27: If you have a faintest shadow of compassion, you should follow the SQL guidelines. 
I beg of you to have mercy on the souls of those reading your code and always start clauses from new lines while writing raw SQL. 
It may not be very important here, but it really is better to get used to the good practices before it's too late and someone — maybe even your future self — gets hurt. 

This applies `SET` clause here and also `WHERE` at $35.

mltdll (changes requested):

app/managers.py
kateryna3k: You're right) Fixed
I'm getting used to this)

mltdll (approved):
GJ.

app/managers.py
```diff
@@ -0,0 +1,40 @@
...
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema_db.db3")
+        self.table_name = "actors"
+
+    def all(self) -> list:
+        actors_cursor = self._connection.execute(
+            f"SELECT * FROM {self.table_name}"
+        )
+        return actors_cursor.fetchall()
```
mltdll, line 13: Sorry, i missed something. This function should return a list of `Actor` instances, not just a list of values.
> - `all` - a method that returns a list of `Actor` instances from DB

mltdll (changes requested):

mltdll (approved):

TPolina (approved):
Good job!

PR #126, opened 2022-10-15
Solution

linaSam:


app/managers.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.db3")
+        self.table_name = "actors"
+
+    def all(self) -> list:
+        actors_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name FROM {self.table_name}"
```
mltdll, line 13: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start `FROM` clause from new line.

mltdll (changes requested):

app/managers.py
linaSam: agree, makes sense

app/managers.py
```diff
@@ -0,0 +1,47 @@
...
+            f"SELECT id, first_name, last_name "
+            f"FROM {self.table_name}"
+        )
+
+        return [Actor(*row) for row in actors_cursor]
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
+            f"(first_name, last_name) VALUES (?, ?)",
```
mltdll, line 22: How about the `VALUES`? :)

mltdll (changes requested):

app/managers.py
linaSam: well, what can i say) give me a minute

mltdll (approved):

app/main.py
```diff
@@ -1,13 +1,13 @@
...
-#     Actor.objects = ActorManager()
-#
-#     Actor.objects.create(first_name="Emma", last_name="Watson")
-#     Actor.objects.create(first_name="Daniel", last_name="Radclife")
-#     print(Actor.objects.all())
-#     Actor.objects.update(2, "Daniel", "Radcliffe")
-#     print(Actor.objects.all())
-#     Actor.objects.delete(1)
-#     print(Actor.objects.all())
+from models import Actor
```
Y-Havryliv, line 1: delete DB from your PR

Y-Havryliv (changes requested):

PR #132, opened 2022-10-16
Solution

Aleksandro777:


app/README.md
```diff
@@ -28,4 +28,4 @@ with given properties.
 #### 4. Test
 Test different methods in the `main.py` module.
 
-### Note: Check your code using this [checklist](checklist.md) before pushing your solution.
+### Note: Check your code using this [checklist](../checklist.md) before pushing your solution.
```
mltdll, line 31: I don't think we should push any changes to this file.

mltdll (changes requested):

mltdll:
What about `README.md`? It seems that you accidentally moved it into `app/`.

mltdll (changes requested):
What about `README.md`? It seems that you accidentally moved it into `app/`.

mltdll (approved):

PR #135, opened 2022-10-17
Solution

VadymShkarbul:


app/managers.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def all(self) -> list[Actor]:
+        actors_rows = self._connection.execute(
+            f"SELECT * FROM {self._table_name}"
+        )
+        return [Actor(*row) for row in actors_rows]
+
+    def update(self, id_upd: int,
+               f_name_upd: str,
+               l_name_upd: str) -> None:
```
mltdll, line 27: These variable names are not very good. If you don't want to use three whole words, you may name them something like `actor_id`, `first_name`, `last_name`.
It's not necessary to duplicate method name in the argument names.

app/managers.py
```diff
@@ -0,0 +1,43 @@
...
+
+        self._connection.execute(
+            f"UPDATE {self._table_name} "
+            "SET first_name = ?, last_name = ?"
+            "WHERE id = ?",
+            (f_name_upd, l_name_upd, id_upd)
+        )
+        self._connection.commit()
+
+    def delete(self, id_del: int) -> None:
```
mltdll, line 37: Same here, I'd suggest to use a name like `id_`, `actor_id`, `id_to_delete`.

mltdll (changes requested):

app/main.py
```diff
@@ -1,13 +1,13 @@
...
-#     Actor.objects = ActorManager()
-#
-#     Actor.objects.create(first_name="Emma", last_name="Watson")
-#     Actor.objects.create(first_name="Daniel", last_name="Radclife")
-#     print(Actor.objects.all())
-#     Actor.objects.update(2, "Daniel", "Radcliffe")
-#     print(Actor.objects.all())
-#     Actor.objects.delete(1)
-#     print(Actor.objects.all())
+from models import Actor
```
Y-Havryliv, line 1: delete DB from your PR

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Nice!

PR #137, opened 2022-10-17
Solution

KatrinLazarenko:


mltdll (approved):

app/managers.py
```diff
@@ -0,0 +1,46 @@
...
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.db3")
+        self.table_name = "cinema"
+
+    def create(self, first_name_: str, last_name_: str) -> None:
```
TPolina, line 11: Why do you add `_` to the end of variable names?

TPolina (changes requested):

app/managers.py
KatrinLazarenko: Danylo shows such approach in video lectures that`s why I wrote in this way.

TPolina: okay, I will ask him)

TPolina (approved):
Good job!

app/managers.py
KatrinLazarenko: I rewatched video. The reason was reserved name, but I missed this mention. Shame on me )

PR #145, opened 2022-10-18
Solution)

Gr1zzZzzly:


app/managers.py
```diff
@@ -0,0 +1,41 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema")
```
Vasyl-Poremchuk, line 8: You forgot to add a file extension, for example `cinema.db3`

Vasyl-Poremchuk (changes requested):

app/managers.py
Gr1zzZzzly: it doesnt work if i change the extension for file

Vasyl-Poremchuk: Okey)

Vasyl-Poremchuk (approved):
Well done!

app/managers.py
TPolina: Why doesn't it work? It should. Probably you configured the extension manually. You can try to delete your db and create a new one. Because adding an extension is a good thing

TPolina (approved):
Good job!

PR #149, opened 2022-10-18
Solution

LiudmylaKulinchenko:


app/managers.py
```diff
@@ -0,0 +1,46 @@
+import sqlite3
```
Vasyl-Poremchuk, line 1: I think you should delete the `cinema.sqlite` file, everything else looks fine)

Vasyl-Poremchuk (changes requested):

Vasyl-Poremchuk (approved):
Well done!

app/managers.py
```diff
@@ -0,0 +1,46 @@
...
+    def update(
+            self, id_to_update: int,
+            first_name: str,
+            last_name: str
+    ) -> None:
+        self._connection.execute(
+            f"UPDATE {self._table_name} "
+            "SET (first_name, last_name) = (?, ?) "
+            "WHERE id = ?",
+            (id_to_update, first_name, last_name,)
```
TPolina, line 35: the order of parameters is wrong

TPolina: did you test your code? it seems to me that the query can be wrong as well

TPolina (changes requested):

TPolina (approved):
Good job!

PR #152, opened 2022-10-18
Solution

fsocie7y:


app/managers.py
```diff
@@ -0,0 +1,44 @@
...
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema")
+        self.table_name = "actors"
+
+    def create(self, first_name_: str, last_name_: str) -> None:
```
mltdll, line 11: I don't see the reason to add `_` to those variables. It's usually done to avoid shadowing names of built-in variables. Do you have any reason?

app/managers.py
```diff
@@ -0,0 +1,44 @@
...
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
```
mltdll, line 13: It's not critical in this case, but I'd still suggest to format raw SQL according to SQL style guides. e.g.
```python
        self._connection.execute(
            f"INSERT INTO {self.table_name} (first_name, last_name) "
            "VALUES (?, ?)",
            (first_name_, last_name_)
        )
```
Same at line $21

mltdll (changes requested):

app/managers.py
fsocie7y: I did this to avoid conflicts with column names in the table

fsocie7y: It's about "_"

mltdll: Wow. But there can be no conflicts there. Those are not only in different namespaces, they are in different programs.

fsocie7y: Okay, I see)

mltdll (approved):

app/managers.py
```diff
@@ -0,0 +1,44 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema")
```
TPolina, line 8: database name should have one of db file formats. "cinema.db3" for example

TPolina (changes requested):

TPolina (approved):
Good job!

PR #153, opened 2022-10-18
'Solution'

Nikyxa:


app/managers.py
```diff
@@ -0,0 +1,52 @@
+import sqlite3
+
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema")
```
Vasyl-Poremchuk, line 9: You forgot to add a file extension, for example `cinema.db3`

app/main.py
```diff
@@ -1,13 +1,12 @@
...
-#     Actor.objects.update(2, "Daniel", "Radcliffe")
-#     print(Actor.objects.all())
-#     Actor.objects.delete(1)
-#     print(Actor.objects.all())
+from models import Actor
+from managers import ActorManager
+
+if __name__ == "__main__":
+    Actor.objects = ActorManager()
+    Actor.objects.create(first_name="Emma", last_name="Watson")
```
Vasyl-Poremchuk, line 6: I think your table will not be created (an error will occur) because you are using parameters named `new_first_name` and `new_last_name` in the `managers.py` module, and there are no such parameters in this module

app/main.py
```diff
@@ -1,13 +1,12 @@
...
+    Actor.objects.create(first_name="Daniel", last_name="Radclife")
```
Vasyl-Poremchuk, line 7: Same here

Vasyl-Poremchuk (changes requested):

app/main.py
Nikyxa: Table was created, but I agree that it would be better to rename parameters)

TPolina (approved):
Good job!

PR #157, opened 2022-10-19
Solution

mariia-movchan:


Mowka7 (approved):
Well done!(〜￣▽￣)〜

app/managers.py
```diff
@@ -0,0 +1,46 @@
...
+            self,
+            id_to_update: int,
+            new_first_name: str,
+            new_last_name: str
+    ) -> None:
+        self._connection.execute(
+            f"UPDATE {self.table_name} "
+            "SET first_name = ?, last_name = ? "
+            "WHERE id = ?",
+            (id_to_update, new_first_name, new_last_name)
```
TPolina, line 36: incorrect order of parameters. Did you test your solution?

TPolina (changes requested):

app/managers.py
mariia-movchan: Omg, thanks, fixed
Sure, but I tested all crud operations at once and I was confident in my solution. So, I don't notice such detail
After all, data was outputed and deleted successfully...

TPolina (approved):
Good job!

PR #159, opened 2022-10-19
Solution

misha-doroshenko:


app/managers.py
```diff
@@ -0,0 +1,46 @@
...
+        select_all = self._connection.execute(
+            f"SELECT * "
+            f"FROM {self.table_name}"
+        )
+        return [Actor(*row) for row in select_all]
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
+            f"(first_name, last_name) VALUES (?, ?) ",
```
mariia-movchan, line 21: extra space at the end of the string
but I think, it doesn't matter

mariia-movchan (approved):
Looks good for me :)

app/managers.py
```diff
@@ -0,0 +1,46 @@
+import sqlite3
+
+from app.models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("../cinema")
```
TPolina, line 8: it's better when your db name has one of db file extensions ( cinema.db3 or cinema.sqlite for example)

app/managers.py
```diff
@@ -0,0 +1,46 @@
...
+        self.table_name = "actor"
+
+    def all(self) -> list:
+        select_all = self._connection.execute(
```
TPolina, line 12: `selct_all` is a bad variable name because it's an action but the variable name should display what is stored in it

TPolina (changes requested):

TPolina (approved):
Good job!

PR #168, opened 2022-11-09
Solution

amirammina:


DmytroSvirsa (changes requested):
manger and database must be in a project root folder, delete cinema.db3 from commit

DmytroSvirsa (approved):

PR #209, opened 2023-01-16
Actor Manager

MarynaProkhorenko:


app/main.py
```diff
@@ -1,13 +1,13 @@
-# from models import Actor
```
olena-hul, line 1: Delete db file from PR

olena-hul (changes requested):

olena-hul:
Good job!

PR #213, opened 2023-01-22
Solution

Phoenix-Erazer:


app/managers.py
```diff
@@ -0,0 +1,50 @@
+from __future__ import annotations
```
olena-hul, line 1: I think it is redundant here

olena-hul (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #215, opened 2023-01-24
solution

Tania-Kharchuk:


app/managers.py
```diff
@@ -0,0 +1,42 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.db3")
+        self.table_name = "actors"
```
Bohdan-Salamakha, line 9: I got an error when tried to test it 
`sqlite3.OperationalError: no such table: actors`
add auto table creating after connection at the __init__. func (don't forget to check if table is already exists)
be aware of users, try to automate what u can

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):
great

PR #216, opened 2023-01-27
'Solution'

artur-leleiko:


app/managers.py
```diff
@@ -0,0 +1,42 @@
+import sqlite3
+
+from app.models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.db3")
```
ArsenPidhoretskyi, line 8: BTW, It is not so good to create a connection here, I recommend moving it to an argument in init with the default value. Check about Dependency Injection

!Optional to change

app/managers.py
```diff
@@ -0,0 +1,42 @@
...
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
+            f"(first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name))
+        self._connection.commit()
+
+    def all(self) -> list:
+        actors_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name "
```
ArsenPidhoretskyi, line 20: When fields will changed `id, first_name, last_name` will be not actual all fields

app/managers.py
```diff
@@ -0,0 +1,42 @@
...
+        self.table_name = "actors"
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
+            f"(first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name))
+        self._connection.commit()
+
+    def all(self) -> list:
```
ArsenPidhoretskyi, line 18: list[Actor] provide full typing

ArsenPidhoretskyi (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #220, opened 2023-01-30
Solution

ArtMot99:


app/managers.py
```diff
@@ -0,0 +1,42 @@
...
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
+            "(first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
+        actors_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name FROM {self.table_name}"
```
ArsenPidhoretskyi, line 20: Better to use SELECT *
Because if you change a Model you need go here and these code also

ArsenPidhoretskyi (changes requested):

PR #222, opened 2023-02-01
'Solution'

Oleksiy-Liubchenko:
Hi, in checklist after was "Make sure you don't push db files (files with .sqlite, .db3, etc. extension)". So I deleted my db file. I needed to do like this or to do something else?


ArsenPidhoretskyi (approved):
Looks good for me

PR #230, opened 2023-02-10
Solution

leetwinoff:


app/managers.py
```diff
@@ -0,0 +1,48 @@
...
+            last_name: str
+    ) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name}"
+            " (first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> None:
```
vasylhnatiuk, line 22: this function return list with data not None, it must be `def all(self) -> list[Actor]:`

vasylhnatiuk (changes requested):

Vasyl-Smutok (approved):
Well Done 👍

PR #239, opened 2023-02-19
solution

musakovskiy:


app/managers.py
```diff
@@ -0,0 +1,39 @@
...
+    def create(self, first_name: str, last_name: str):
+        self._connection.execute(
+            f'INSERT INTO {self.table_name} (first_name, last_name) VALUES (?)',
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self):
+        actors_cursor = self._connection.execute(
+            f'SELECT id, first_name, last_name FROM {self.table_name}'
```
ArsenPidhoretskyi: Use select * 
Because when you change your actor class you also must to edit code here

app/models.py
```diff
@@ -0,0 +1,11 @@
...
+
+from app.managers import ActorManager
+
+
+@dataclass
+class Actor:
+    objects = ActorManager()
+    id: int
+    first_name: str
+    last_name: str
```
ArsenPidhoretskyi, line 11: Change order
```suggestion
    id: int
    first_name: str
    last_name: str
    
    objects = ActorManager()
```

ArsenPidhoretskyi (changes requested):

app/managers.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name}"
+            f"(first_name, last_name) VALUES (?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
```
olena-hul, line 19:
```suggestion
    def all(self) -> list[Actor]:
```

olena-hul (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #245, opened 2023-02-25
Solution

SafonovVladimir:
add my solution

### DIFF
--- a/app/managers.py
+++ b/app/managers.py
@@ -1,0 +1,47 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema_db.db3")
+        self.table_name = "actors"
+
+    def all(self) -> list[Actor]:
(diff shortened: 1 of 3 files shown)


ZabFTFT (approved):
Well done!

app/managers.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema_db.db3")
+        self.table_name = "actors"
+
+    def all(self) -> list[Actor]:
+        actors_coursor = self._connection.execute(
+            "SELECT * FROM actors"
```
Y-Havryliv, line 13: you have `{self.table_name}` everywhere but you used `actors` here

Y-Havryliv (changes requested):

app/managers.py
SafonovVladimir: after adding self.table_name I forgot to change this.
Thank's, fixed

Y-Havryliv (approved):
GJ!

PR #246, opened 2023-02-25
Solution

SlavikSherbak:


app/models.py
```diff
@@ -0,0 +1,8 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class Actors:
```
katevasilenko, line 5: We named classes in Python in singular. And also please fix all over the project
```suggestion
class Actor:
```

katevasilenko (changes requested):

Grayder0152 (approved):

PR #253, opened 2023-02-28
Solution

ZabFTFT:


WDemidenko (approved):
Good job!

app/managers.py
```diff
@@ -0,0 +1,48 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect(
+            "C:\\Programming Study\\Mate\\py-actor-manager\\cinema.db3"
```
Y-Havryliv, line 9: it would be easier and you could avoid using the absolute path by moving your db into the app directory

Y-Havryliv (approved):

app/managers.py
ZabFTFT: I see, thanks!

PR #262, opened 2023-03-03
solution

avkpol:


app/managers.py
```diff
@@ -0,0 +1,45 @@
...
+
+class ActorManager:
+
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.db3")
+        self.table_name = "actors"
+
+    def all(self) -> None:
+        actors_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name FROM {self.table_name}"
```
ArsenPidhoretskyi, line 14: Use select *, because when you update your Actor model you must change sql also

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #268, opened 2023-03-06
solution

DavidGlusch:


BohdanLazaryshyn (approved):

Serhii-Leonenko (changes requested):
Read the checklist again!

app/managers.py
```diff
@@ -0,0 +1,48 @@
+import sqlite3
+from app.models import Actor
```
wQuelS, line 2: follow import style according to PEP8
https://peps.python.org/pep-0008/#imports

wQuelS (changes requested):

app/managers.py
DavidGlusch: as i understood, you were refering to this part

wQuelS (approved):
Great

PR #272, opened 2023-03-06
Solution

ZhAlexR:


amber-marichi (approved):
nice

app/main.py
```diff
@@ -1,13 +1,13 @@
...
-#     print(Actor.objects.all())
-#     Actor.objects.update(2, "Daniel", "Radcliffe")
-#     print(Actor.objects.all())
-#     Actor.objects.delete(1)
-#     print(Actor.objects.all())
+from models import Actor
+from managers import ActorManager
+
+if __name__ == "__main__":
+    Actor.objects = ActorManager("cinema_db.db3")
```
Serhii-Leonenko, line 5: According to the task the name of the db should be `cinema`, but it's not a big problem

Serhii-Leonenko (approved):

app/main.py
ZhAlexR: > 

I supposed that adding  "db" to the name of file in the end is a good practice.
Thanks! Will use simple naming variant next time)

PR #278, opened 2023-03-07
Solution DonchykD

DmytryTong:


Yulia-Mala (approved):
:)

app/managers.py
```diff
@@ -0,0 +1,44 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("Actor_db.sqlite")
```
Serhii-Leonenko, line 8: It's better to use .db3 file extension which is associated with SQLite version 3

Serhii-Leonenko (changes requested):

app/managers.py
DmytryTong: Done.

wQuelS (approved):
GJ

PR #280, opened 2023-03-07
'Solution'

Morty67:


app/managers.py
```diff
@@ -0,0 +1,40 @@
+import sqlite3
+from models import Actor
```
Anatolii-Poznyak, line 2: pep8 : D

Anatolii-Poznyak (approved):
Well Done !

app/managers.py
Morty67: what wrong with pep8? in PyCharm i don't have any problems

Serhii-Leonenko: https://peps.python.org/pep-0008/#imports

Serhii-Leonenko (changes requested):

app/managers.py
Morty67: > https://peps.python.org/pep-0008/#imports

ty, i read

Serhii-Leonenko (approved):
Well done!

PR #282, opened 2023-03-07
Solution

TeliksPG:


ghost (approved):

app/managers.py
```diff
@@ -0,0 +1,40 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.db")
```
Serhii-Leonenko, line 8: The difference between .db and .db3 file extensions is that .db files are associated with the older SQLite version 2, while .db3 files are associated with the SQLite version 3

Serhii-Leonenko (changes requested):

app/managers.py
ProdyRodion: Danya said, that there is no difference, and in SQLite 3 i don't have any other option

wQuelS (approved):
Good

PR #285, opened 2023-03-07
Solution

Anatolii-Poznyak:


app/managers.py
```diff
@@ -0,0 +1,50 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.sqlite")
```
wQuelS, line 8: It's better to use .db3 file extension which is associated with SQLite version 3

app/managers.py
```diff
@@ -0,0 +1,50 @@
...
+        self.table_name = "actors"
+
+    def all(self) -> list:
+        return [
+            Actor(*actor)
+            for actor
+            in self._connection.execute(f"SELECT * FROM {self.table_name}")
```
wQuelS, line 15: Can you first make this execution into a separate variable, and after that make this list comprehension?

wQuelS (changes requested):

app/main.py
```diff
@@ -1,6 +1,6 @@
 # from models import Actor
 # from managers import ActorManager
-#
+# #
```
wQuelS, line 3: just uncomment this whole file

wQuelS (changes requested):

app/managers.py
```diff
@@ -0,0 +1,51 @@
...
+            (new_first_name, new_last_name, id_for_update)
+        )
+        self._connection.commit()
+
+    def delete(
+            self,
+            id_to_delete: int
+    ) -> None:
+        self._connection.execute(
+            f"DELETE FROM {self.table_name} WHERE id = ?",
```
YuriiKindrat, line 48: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start WHERE clause from new line.

YuriiKindrat (changes requested):

PR #286, opened 2023-03-07
Solution

Litvinenko23:


ostapT (approved):
GJ!)

app/managers.py
```diff
@@ -0,0 +1,45 @@
+import sqlite3
+from typing import List
+
+from app.models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.sqlite")
```
wQuelS, line 9: It would be better to use .db3 file extension which is associated with SQLite version 3

wQuelS (changes requested):

wQuelS (approved):
Good

PR #288, opened 2023-03-07
Solution

ghost:


Yevheniy-Lototskiy (approved):

app/managers.py
```diff
@@ -0,0 +1,45 @@
+import sqlite3
+
+from app.models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.sqlite")
```
wQuelS, line 8: It's better to use .db3 file extension which is associated with SQLite version 3

wQuelS (changes requested):

YuriiKindrat (approved):

PR #289, opened 2023-03-07
'Solution'

alexlega:


inrock1 (approved):
GJ

app/managers.py
```diff
@@ -0,0 +1,43 @@
+import sqlite3
+from models import Actor
```
wQuelS, line 2: make it according to PEP8
https://peps.python.org/pep-0008/#imports

app/managers.py
```diff
@@ -0,0 +1,43 @@
+import sqlite3
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema")
```
wQuelS, line 7: what about extension?
Hint: .db extension is usually associated with SQLite vers2, while .db3 with SQLite 3.

wQuelS (changes requested):

app/managers.py
alexlega: Thanks for advice! I didn't have any extension in pycharm or in a directory, so I decided that it should work well like this

wQuelS (approved):
Great

PR #291, opened 2023-03-07
"solution"

skmiwa1:


myroslavatrschk (changes requested):
You forget to add the "managers.py" module in your commit.

app/managers.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
+            f"(first_name, last_name) VALUES(?,?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> None:
```
Y-Havryliv, line 21: does it return None?

app/managers.py
```diff
@@ -0,0 +1,44 @@
...
+        data = self._connection.execute(
+            f"SELECT * From {self.table_name}"
```
Y-Havryliv, line 23: can you write it in a better way? `From`

Y-Havryliv (changes requested):

app/managers.py
skmiwa1: it should return list!

app/managers.py
```diff
@@ -0,0 +1,44 @@
...
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
+            f"(first_name, last_name) VALUES(?,?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
+        data = self._connection.execute(
+            f"SELECT * FROM {self.table_name}"
```
YuriiKindrat, line 23: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM clause from new line.

YuriiKindrat (changes requested):

Y-Havryliv (approved):
Nice!

PR #292, opened 2023-03-07
Solution

ada-krav:


Zenionoi (approved):

app/managers.py
```diff
@@ -0,0 +1,41 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema")
```
YuriiKindrat, line 8: it is better to use file extension, for example cinema.db3

YuriiKindrat (changes requested):

wQuelS (approved):
Good

PR #296, opened 2023-03-07
Solution

Yevheniy-Lototskiy:


app/managers.py
```diff
@@ -0,0 +1,45 @@
...
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
+            f"(first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
+        actors_data_cursor = self._connection.execute(
+            f"SELECT * FROM {self.table_name}"
```
YuriiKindrat, line 21: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM clause from new line.

app/managers.py
```diff
@@ -0,0 +1,45 @@
...
+            f"UPDATE {self.table_name} "
+            "SET first_name = ?, last_name = ? "
+            "WHERE id = ?",
+            (new_first_name, new_last_name, actor_id)
+        )
+        self._connection.commit()
+
+    def delete(self, id_to_delete: int) -> None:
+        self._connection.execute(
+            f"DELETE FROM {self.table_name} "
```
YuriiKindrat, line 41: same

YuriiKindrat (changes requested):

app/managers.py
```diff
@@ -0,0 +1,47 @@
...
+            f"FROM {self.table_name}"
+        )
+
+        return [Actor(*row) for row in actors_data_cursor]
+
+    def update(
+            self, actor_id: int,
+            new_first_name: str,
+            new_last_name: str
+    ) -> None:
```
Serhii-Leonenko, line 31: Fix indentation here

app/managers.py
```diff
@@ -0,0 +1,47 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.sqlite")
```
Serhii-Leonenko, line 8: It's better to use here `.db3` file extension

Serhii-Leonenko (changes requested):

Serhii-Leonenko (changes requested):
Also remove db file from your PR

app/managers.py
Yevheniy-Lototskiy: Can I use .sqlite3 instead .db3?

PR #298, opened 2023-03-07
'Solution'

oksanaaam:


app/managers.py
```diff
@@ -0,0 +1,42 @@
...
+    def create(self, format_: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} (format) VALUES (?)",
+            (format_,)
+        )
+        self._connection.commit()
+
+    def all(self) -> List[Actor]:
+        actor_managers_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name FROM {self.table_name}"
```
YuriiKindrat, line 22: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM clause from new line.

app/managers.py
```diff
@@ -0,0 +1,42 @@
...
+            f"UPDATE {self.table_name} "
+            "SET format = ? "
+            "WHERE id = ?",
+            (new_format, id_to_update)
+        )
+        self._connection.commit()
+
+    def delete(self, id_to_delete: int) -> None:
+        self._connection.execute(
+            f"DELETE FROM {self.table_name} "
```
YuriiKindrat, line 38: same

YuriiKindrat (changes requested):

app/managers.py
```diff
@@ -0,0 +1,44 @@
+import sqlite3
+
+from typing import List
```
Serhii-Leonenko, line 3: Do you need a blank line between these imports?

Serhii-Leonenko (changes requested):

wQuelS (approved):
Good

PR #300, opened 2023-03-07
Solution

Zenionoi:


app/managers.py
```diff
@@ -0,0 +1,34 @@
...
+        return actors
+
+    def update(self, id: int, first_name: str, last_name: str) -> None:
+        self.cursor.execute("UPDATE actors SET first_name = ?,"
+                            " last_name = ? WHERE id = ?",
+                            (first_name, last_name, id))
+        self.connect.commit()
+
+    def delete(self, id: int) -> None:
+        self.cursor.execute("DELETE FROM actors WHERE id = ?", (id,))
```
YuriiKindrat, line 33: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM/WHERE etc. clause from new line.

YuriiKindrat (changes requested):

app/managers.py
```diff
@@ -0,0 +1,36 @@
...
+                            "FROM actors")
+        actors = []
+        for row in self.cursor.fetchall():
+            actor = Actor(*row)
+            actors.append(actor)
+        return actors
+
+    def update(self, id: int, first_name: str, last_name: str) -> None:
+        self.cursor.execute("UPDATE actors SET first_name = ?, last_name = ? "
+                            "WHERE id = ?",
```
wQuelS, line 29: I guess `SET` should be on a separate line as well.

wQuelS (changes requested):
And where is your `main.py` file?

YuriiKindrat (approved):

PR #305, opened 2023-03-09
Solution

6aPa6aH:


app/managers.py
```diff
@@ -0,0 +1,48 @@
...
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
+            f"(first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
+        actors_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name "
```
ArsenPidhoretskyi, line 22: Use select *, because if you change Actor model you also need to change here

ArsenPidhoretskyi (changes requested):

app/managers.py
```diff
@@ -0,0 +1,47 @@
+import sqlite3
+from models import Actor
```
katevasilenko, line 2: You need the blank line between third-party and your own imports

katevasilenko (changes requested):
Only one neat-picky remark, all other looks good

vasylhnatiuk (approved):
Good job!

PR #312, opened 2023-03-13
Solution

StreetNik:


app/managers.py
```diff
@@ -0,0 +1,46 @@
+import sqlite3
+from models import Actor
+
+from typing import List
+
+
+class ActorManager:
```
Grayder0152, line 7: pls add method for closing connection

Grayder0152 (changes requested):

Grayder0152 (approved):

PR #313, opened 2023-03-15
Solution

xtina999:


app/managers.py
```diff
@@ -0,0 +1,48 @@
...
+            f"INSERT INTO {self.table_name} "
+            f"(first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    # READ
+    def all(self) -> list:
+        actor_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name FROM {self.table_name} "
```
ArsenPidhoretskyi, line 22: Better to use select *, because if you change actor model you need also change here

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Well done!

PR #321, opened 2023-03-19
Development

NVMBR13:


app/managers.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.db3")
+        self.table = "actors"
+
+    def all(self) -> list[Actor]:
+        actor_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name FROM {self.table}"
```
ArsenPidhoretskyi, line 13: Use select *
Because if you change actor model you need to change also this

app/managers.py
```diff
@@ -0,0 +1,39 @@
...
+        )
+        return [Actor(actor[0], actor[1], actor[2]) for actor in actor_cursor]
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
```
ArsenPidhoretskyi, line 18: Actor(*actor)

ArsenPidhoretskyi (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #324, opened 2023-03-24
'Solution'

Mykhailo-Benzel:


app/managers.py
```diff
@@ -0,0 +1,46 @@
...
+        self._connection.execute(
+            f"INSERT INTO {self.table_name}"
+            " (first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name,)
+        )
+        self._connection.commit()
+
+    def all(self) -> List[Actor]:
+        actor_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name FROM {self.table_name}"
```
ArsenPidhoretskyi, line 22: In this case better to use: SELECT * 
Because when you change your Actor model you also need go and change code here

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Good job!

PR #328, opened 2023-03-29
Solution

yuraua5:


app/managers.py
```diff
@@ -0,0 +1,36 @@
...
+from app.models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.db3")
+        self.table_name = "actors"
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(f"INSERT INTO {self.table_name}"
```
vasylhnatiuk, line 12:
```suggestion
        self._connection.execute(
            f"INSERT INTO {self.table_name} "
            f"(first_name, last_name) "
            f"VALUES (?, ?)",
            (first_name, last_name,)
        )

```

app/managers.py
```diff
@@ -0,0 +1,36 @@
...
+                                 f" (first_name, last_name)"
+                                 f" VALUES (?, ?)",
+                                 (first_name, last_name))
+        self._connection.commit()
+
+    def all(self) -> list[Actor]:
+        actors_actor_cursor = self._connection.execute("SELECT * FROM actors")
```
vasylhnatiuk, line 19:
```suggestion
        actors_actor_cursor = self._connection.execute(f"SELECT * FROM {self.table_name}")

``` it would be better to use

app/managers.py
vasylhnatiuk: the same in all of the code

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
Do it

ArsenPidhoretskyi (approved):
Well done!

PR #330, opened 2023-03-29
solution

MkrtychKhachatrian:


app/managers.py
```diff
@@ -0,0 +1,45 @@
...
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
+            "(first_name, last_name) VALUES(?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> List[Actor]:
+        actors_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name FROM {self.table_name}"
```
vasylhnatiuk, line 22: what if will be new field?
```suggestion
            f"SELECT * FROM {self.table_name}"
```

vasylhnatiuk (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #336, opened 2023-04-08
Solution

Roman28101:


app/managers.py
```diff
@@ -0,0 +1,45 @@
...
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
+            "(first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
+        actors_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name  FROM {self.table_name}"
```
vasylhnatiuk, line 20:
```suggestion
            f"SELECT *  FROM {self.table_name}"
```

vasylhnatiuk: you need `all `fields

vasylhnatiuk (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #337, opened 2023-04-17
Develop

andy77andy:


Vasyl-Smutok (changes requested):
remove DB from PR 
image

Vasyl-Smutok (changes requested):
also remove identifier.sqlite

image

app/managers.py
```diff
@@ -0,0 +1,34 @@
...
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self.table_name = "actors"
+        self._connection = sqlite3.connect("cinema")
+
+    def all(self) -> list:
+        data = self._connection.execute("SELECT id, first_name, last_name "
```
ArsenPidhoretskyi, line 12: Better to use select * because when you will change Actor model. You must change this code also

app/models.py
```diff
@@ -0,0 +1,8 @@
+from dataclasses import dataclass
+
+
+@dataclass()
```
ArsenPidhoretskyi, line 4:
```suggestion
@dataclass

```

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #341, opened 2023-04-19
Solution

Anyta17:


app/managers.py
```diff
@@ -0,0 +1,44 @@
...
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
+            f"(first_name, last_name) VALUES (?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
+        actors_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name FROM {self.table_name}"
```
ArsenPidhoretskyi, line 20: Better to do SELECT *. Because for now when you change Actor model you also need to change this code

ArsenPidhoretskyi (changes requested):

Bohdan-Salamakha (approved):

PR #345, opened 2023-05-01
Solution

DmytroBondariev:


app/managers.py
```diff
@@ -0,0 +1,51 @@
...
+
+@dataclass
+class ActorManager:
+    _connection = sqlite3.connect("cinema.db3")
+    table_name = "actors"
+
+    def all(self) -> list:
+        cinema_cursor = self._connection.execute(
+            "SELECT id, first_name, "
+            f"last_name FROM {self.table_name}"
```
Vasyl-Smutok, line 15: Use SELECT *
Because if you will change the Actor model you also must change this part of code

app/managers.py
```diff
@@ -0,0 +1,51 @@
...
+
+from models import Actor
+
+
+@dataclass
+class ActorManager:
+    _connection = sqlite3.connect("cinema.db3")
+    table_name = "actors"
+
+    def all(self) -> list:
```
Vasyl-Smutok, line 12:
```suggestion
    def all(self) -> list[Actor]:

```

Vasyl-Smutok (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #347, opened 2023-05-01
solution

VladyslavOladko:


app/managers.py
```diff
@@ -0,0 +1,32 @@
...
+        self.table_name = "actors"
+
+    def create(self, first_name: str, last_name: str):
+        self._connection.execute(f"INSERT INTO {self.table_name} (first_name, last_name) VALUES (?, ?)",
+                                 (first_name, last_name)
+                                 )
+        self._connection.commit()
+
+    def all(self):
+        actor_cursor = self._connection.execute(f"SELECT id, first_name, last_name FROM {self.table_name}")
```
ArsenPidhoretskyi, line 17: Use select * because when you change Actor model you also must change Manager. And it will be unlogical

ArsenPidhoretskyi (changes requested):
Use flake8. Check if line length < 80

app/managers.py
```diff
@@ -0,0 +1,41 @@
+import sqlite3
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("D:/for prog/library_db/cinema.db3")
```
ArsenPidhoretskyi, line 7: No need to provide full path. Provide relative path:
cinema.db3

ArsenPidhoretskyi (changes requested):

app/managers.py
```diff
@@ -0,0 +1,41 @@
...
+                                 f"(first_name, last_name) "
+                                 f"VALUES (?, ?)",
+                                 (first_name, last_name)
+                                 )
+        self._connection.commit()
+
+    def all(self) -> list:
+        actor_cursor = self._connection.execute(f"SELECT id, "
+                                                f"first_name, "
+                                                f"last_name "
```
ArsenPidhoretskyi, line 21: Use select *, because if you will change Actor model you also need change here

app/managers.py
```diff
@@ -0,0 +1,41 @@
...
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.db3")
+        self.table_name = "actors"
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(f"INSERT INTO {self.table_name} "
+                                 f"(first_name, last_name) "
+                                 f"VALUES (?, ?)",
+                                 (first_name, last_name)
+                                 )
```
ArsenPidhoretskyi, line 15:
```suggestion
        self._connection.execute(
            f"INSERT INTO {self.table_name} (first_name, last_name) "
            f"VALUES (?, ?)",
            (first_name, last_name)
        )
```

ArsenPidhoretskyi: Format the same in other places

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Great 👍

PR #351, opened 2023-05-07
Solution

goldenuni:


app/managers.py
```diff
@@ -0,0 +1,37 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema_db")
```
Bohdan-Salamakha, line 8: db should have .sqlite3 extension

Bohdan-Salamakha (changes requested):
u need to create all required tables automatically
image

app/managers.py
goldenuni: @Bohdan-Salamakha 
image
As I understand, I should create DB inside SQLiteStudio and then add it into a project, isn't it? If yes, I've done this, but when I add extension for the file, my program doesn't see the DB. To resolve your problem, I should have added the DB file into Git, but a checklist said, that I [shouldn't](https://github.com/mate-academy/py-actor-manager/blob/master/checklist.md#dont-push-db-files). So what do I need to do?

ArsenPidhoretskyi: cinema_db is a file it does not SQL database. Please read about extensions of SQLite
You should not add db file into git. Read about gitignore

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #353, opened 2023-05-08
'Solution'

HalynaPetrova:


app/managers.py
```diff
@@ -0,0 +1,47 @@
+import sqlite3
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema")
```
Grayder0152, line 7: Where do you close connection ?

Grayder0152 (changes requested):

app/managers.py
HalynaPetrova: automatically? by analogy with the context manager?

app/managers.py
```diff
@@ -0,0 +1,47 @@
...
+        self.table_name = "actors"
+
+    def all(self) -> list:
+        actors_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name "
```
vasylhnatiuk, line 12:
```suggestion
            f"SELECT * "
```

vasylhnatiuk, line 12:
```suggestion
            f"SELECT * "
```

vasylhnatiuk (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #356, opened 2023-05-12
Solution of actor manager

svyshcho:


app/managers.py
```diff
@@ -0,0 +1,45 @@
...
+            f"INSERT INTO {self.table_name}"
+            f" (first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    # read
+    def all(self) -> list[Actor]:
+        actors_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name FROM {self.table_name}"
```
ArsenPidhoretskyi, line 22: Use select * for select all fields

ArsenPidhoretskyi (changes requested):

app/managers.py
svyshcho: Done

olena-hul:
Good job!

PR #380, opened 2023-06-09
Initial implementation

thorium-zero:


app/managers.py
```diff
@@ -0,0 +1,46 @@
...
+        self.__connector: Connection = connect(db_path)
+        self.__actor_helper: EntityHelper = EntityHelper(Actor, "actors")
+
+    def __del__(self) -> None:
+        if self.__connector:
+            self.__connector.close()
+
+    def create(self, actor: Actor) -> None:
+        cursor = self.__connector.execute(
+            self.__actor_helper.get_insert_statement_param(),
```
LiudmylaKulinchenko, line 19: Why don't you want to put f-string here directly instead of calling a function, which will just give you an attribute of __actor_helper?

app/entity_helper.py
```diff
@@ -0,0 +1,84 @@
...
+class EntityHelper:
+    __placeholder_symbol = "?"
+
+    def __init__(self, entity_type: type, table_name: str) -> None:
+        if not is_dataclass(entity_type):
+            raise ValueError(f"EntityHelper is able to work with "
+                             f"dataclasses only. {type} is not a dataclass")
+        self.__entity_type: type = entity_type
+        self.__table_name: str = table_name
+        self.__entity_type_fields: tuple = tuple(
```
LiudmylaKulinchenko, line 16: no need to annotate types for variables if you understand what will be in them from their initialization

LiudmylaKulinchenko (changes requested):

app/entity_helper.py
```diff
@@ -0,0 +1,84 @@
...
+            .__dataclass_fields__
+            .keys()
+        )
+        self.__insert_into_statement: str = (
+            f"insert into '{self.__table_name}' "
+            f"{self.__entity_type_fields[1:]} values "
+            f"({self.make_placeholders(len(self.__entity_type_fields) - 1)})"
+        )
+        self.__select_all_statement: str = (
+            f"select * from {self.__table_name}"
```
LiudmylaKulinchenko, line 27: Using uppercase for SQL commands (e.g., SELECT, INSERT, UPDATE) has been a common convention and is often recommended in many coding style guides. This convention helps differentiate SQL keywords from table and column names, making the code more readable and easier to understand

app/managers.py
thorium-zero: @LiudmylaKulinchenko  Because I want to stay consistent in my mapper implementation. I could create properties for almost all of those statements. All, but the update statement, since it can not simply be parametrized with a template just like other sql commands. I don't like an idea of making some of object's attributes accessible by properties and others - by methods. The only reasonable solution of your request here is to implement another layer of abstraction so I could hide those implementation details withing some sort of a `data set` class as a Repository/Unit of work subsystem, and that is something that is far outside of a scope of responsibilities for that EntityHelper class, which is, essentially, a low level mapper. 

All in all, I want to keep this design as it is

app/entity_helper.py
thorium-zero: I am aware of that, but I don't want to depend on the `__init__`'s parameters type hints when I analyze my code. I've seen no rules that prevent me from doing this kind of hinting, so I tend to do it, at least for my fields inside of my classes. It is easier to read the code this way, so until there are no company rules against this style, I want to follow it

app/entity_helper.py
thorium-zero: yeah, I know, I'm just too old to do it like this 🤪 I do use this approach of using uppercase for sql commands when I when I write a separate *.sql files, but polluting my source code with UPPERCASE - it is something I will avoid, unless it's a company policy to write it like this. Uppercased constant values, `defined` (well-known) literals and sql statements - it is something that makes our code smell. So, yeah, I never use this convention inside my non-sql files for that reason

PR #392, opened 2023-06-23
'Solution'

V-Shkrobatskyi:


app/managers.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.db3")
+        self.table_name = "actors"
+
+    def all(self) -> [Actor]:
+        actors_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name FROM {self.table_name}"
```
vsmutok, line 13: Use select * for selecting all fields

vsmutok (changes requested):
Also remove DB from PR

app/managers.py
V-Shkrobatskyi: what is PR?

V-Shkrobatskyi: just remove DB

ArsenPidhoretskyi (approved):
Well done!

PR #406, opened 2023-06-27
solution

oleksiikolii:


kovaliskoveronika (approved):

app/managers.py
```diff
@@ -0,0 +1,35 @@
+import sqlite3
+from models import Actor
```

app/managers.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.db")
+        self._table_name = "actors"
+
+    def all(self) -> list:
+        sql = (f"SELECT id, first_name, last_name "
```
LiudmylaKulinchenko, line 11: You can skip creating additional variable in every function. But if you thinks it looks better, it is ok

LiudmylaKulinchenko (changes requested):

app/managers.py
LiudmylaKulinchenko: here should be only one empty line)

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🌼🎈

PR #412, opened 2023-06-27
Solution

volodymyr-komarnyckyi:


mariiahorbova (approved):

app/managers.py
```diff
@@ -0,0 +1,43 @@
+import sqlite3
+from app.models import Actor
```

LiudmylaKulinchenko (approved):
Good job! 🌼

PR #414, opened 2023-06-28
done

maccarets:


app/managers.py
```diff
@@ -0,0 +1,38 @@
+import sqlite3
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self.conn = sqlite3.connect("cinema.db3")
```
oleksiikolii, line 7: It's better name it connection, no need to short

oleksiikolii (approved):

app/managers.py
```diff
@@ -0,0 +1,38 @@
+import sqlite3
+from models import Actor
```

app/managers.py
```diff
@@ -0,0 +1,38 @@
...
+        self.connection.commit()
+
+    def delete(self, id: int) -> None:
+        self.connection.execute(f"DELETE FROM {self.table_name} "
+                                f"WHERE id = ?",
+                                (id,))
+        self.connection.commit()
+
+
+if __name__ == "__main__":
```
LiudmylaKulinchenko, line 34: remove `if __name__ == "__main__":`

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):

PR #417, opened 2023-06-28
Solution

Po-SvinSkipodla:


app/managers.py
```diff
@@ -0,0 +1,45 @@
+import sqlite3
+from app.models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("../cinema.sqlite")
+
```
IhorVoskoboinikov, line 8: Add to __init__ table name and use it in your SQL requests.

IhorVoskoboinikov (changes requested):
Add one upgrade to your code)

app/managers.py
Po-SvinSkipodla: i think it's easier to write actors than make f strings with self.table_name. Otherwords i disagree with this

LiudmylaKulinchenko: Using the self.table_name attribute allows you to easily manage and reuse the table name in multiple SQL queries, making your code more organized and maintainable. What will you do if you'll need to change a table name?
Refactoring the whole file is pretty much worse than change talbe name in one place

app/managers.py
```diff
@@ -0,0 +1,45 @@
+import sqlite3
+from app.models import Actor
```
LiudmylaKulinchenko, line 2: add an empty line between third-party and local imports

LiudmylaKulinchenko (changes requested):

Oleg-s-s (approved):
:+1:

PR #425, opened 2023-07-07
Solution

NazarLytvyniuk:


app/managers/manager.py
```diff
@@ -0,0 +1,47 @@
+import sqlite3
+
+from app.models.actor import Actor
+
+
+class ActorManager:
+
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema")
```
Bohdan-Salamakha, line 9: the database should have an extension like .sqlite3

app/managers/manager.py
```diff
@@ -0,0 +1,47 @@
...
+            f"FROM {self.table_name}"
+        )
+
+        return [Actor(*row) for row in actor_cursor]
+
+    def update(
+            self,
+            id_to_update: int,
+            new_first_name: str,
+            new_last_name: str) -> None:
```
Bohdan-Salamakha, line 32:
```suggestion
    def update(
            self,
            id_to_update: int,
            new_first_name: str,
            new_last_name: str
            ) -> None:
```

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #443, opened 2023-08-02
Solution actor manager

MarkShapka:


app/managers.py
```diff
@@ -0,0 +1,36 @@
...
+from app.models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.db3")
+
+    def all(self) -> None:
+        actors_cursor = self._connection.execute(
+            "SELECT id, first_name, last_name FROM actors"
```
pavlejviki, line 12: You can use * to select all the columns from the table.

It is better to create an attribute to store the table name. By defining the table name as an attribute, you make the code more flexible and easier to update if the table name ever needs to be changed.

pavlejviki (changes requested):

app/managers.py
```diff
@@ -0,0 +1,36 @@
+import sqlite3
+
+from app.models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.db3")
+
```
pavlejviki, line 9: You haven't fixed the previous comment regarding the table name.

Create an attribute to store the table name  and use it throughout your code.


```suggestion
self.table_name = "actors"
```

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #452, opened 2023-08-18
'Solution'

niksgawrilenko:


app/managers.py
```diff
@@ -0,0 +1,45 @@
...
+        )
+        self._connect.commit()
+
+    def all(self) -> List[Actor]:
+        actor_cursor = self._connect.execute(
+            f"SELECT * FROM {self.table_name}"
+        )
+        return [Actor(*row) for row in actor_cursor]
+
+    def update(self,
```
Polyakiv-Andrey, line 27: After opening the parentheses, write all the arguments on a new line, including the first one, the closing bracket is also written on a new line

Polyakiv-Andrey (changes requested):

app/managers.py
niksgawrilenko: done

PR #459, opened 2023-08-31
Develop

VasylTurok:


cinema
Dimosphen1, line 1: Remove this binary file from your PR, you could optionally leave it in the app directory but it would be better to not include a DB file.

app/managers.py
```diff
@@ -0,0 +1,59 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._conection = sqlite3.connect("../cinema")
```
Dimosphen1, line 8: Create a connection assuming that db is in the current working directory and has a `db3` extension.

app/managers.py
```diff
@@ -0,0 +1,59 @@
...
+    def delete(self, id_actor: int) -> None:
+        self._conection.execute(
+            f"DELETE FROM {self.table_name} "
+            f"WHERE id = ?",
+            (id_actor, )
+        )
+        self._conection.commit()
+
+    def print_(self) -> None:
+        print(self._conection)
```
Dimosphen1, line 53: There is no need to add this method here, remove it from your class.

app/managers.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+if __name__ == "__main__":
+    am = ActorManager()
+    am.update(1, "Sania", "Sanich ")
+    print(am.print_())
```
Dimosphen1, line 59: Remove this code from your PR, it is not needed.

Dimosphen1 (changes requested):
Several changes were requested.

PR #466, opened 2023-09-05
'Solution'

lgSilay:


OlegatorLE (approved):
GJ!

app/managers.py
```diff
@@ -0,0 +1,42 @@
+import sqlite3
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema_db.db3")
+        self.table_name = "actors"
+
```
Sofbog, line 9: Before insert data in db we need to make sure table is existing or we need to create it

app/managers.py
```diff
@@ -0,0 +1,42 @@
...
+            self,
+            id_to_update: int,
+            new_first_name: str,
+            new_last_name: str
+    ) -> None:
+        self._connection.execute(
+            f"UPDATE {self.table_name} "
+            "SET first_name = ?, last_name = ?"
+            "WHERE id = ?",
+            (new_first_name, new_last_name, id_to_update)
```
Sofbog, line 33: ncorrect order of parameters, so update doesn't work correctly.

Sofbog (changes requested):

Sofbog (approved):
Well Done

PR #472, opened 2023-09-06
Solution

anna-lybid:


app/managers.py
```diff
@@ -0,0 +1,54 @@
...
+        self._connection = sqlite3.connect("cinema.sqlite")
+
+        with self._connection:
+            self._connection.execute(
+                f"CREATE TABLE IF NOT EXISTS {self.table_name} ("
+                "id INTEGER PRIMARY KEY, "
+                "first_name VARCHAR(50) NOT NULL, "
+                "last_name VARCHAR(50) NOT NULL"
+                ")"
+            )
```
ORENYT, line 18: I thought we have to create the table from DB sql window.

app/managers.py
```diff
@@ -0,0 +1,54 @@
...
+            f"INSERT INTO {self.table_name} "
+            f"(first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name))
+        self._connection.commit()
+
+    def all(self) -> list[Actor]:
+        actors = self._connection.execute(
+            f"SELECT * FROM {self.table_name}"
+        )
+        return [Actor(*row) for row in actors]
```
ORENYT, line 32: good idea to use unpacking

ORENYT (approved):
Nice solution

app/managers.py
Kroll410: @ORENYT  It wasn't mentioned

PR #487, opened 2023-09-06
Solution

CaCtUs1k:


obohatov (approved):
Not sure if the database file should have been sent

app/managers.py
```diff
@@ -0,0 +1,46 @@
...
+            "Where id = ?",
+            (new_first_name, new_last_name, id_to_upd)
+        )
+        self._connection.commit()
+
+    def delete(self, id_to_del: int) -> None:
+        self._connection.execute(
+            f"DELETE FROM {self.table_name} "
+            "WHERE id = ?",
+            (id_to_del,)
```
Sofbog, line 44: I don't think` id_to_del` is a very good abbreviation. It only saves you two characters anyway. So why don't you rename the variable into `id_to_delete`

app/managers.py
```diff
@@ -0,0 +1,46 @@
...
+            self,
+            id_to_upd: int,
+            new_first_name: str,
+            new_last_name: str
+    ) -> None:
+        self._connection.execute(
+            f"UPDATE {self.table_name} "
+            "SET first_name = ?, last_name = ? "
+            "Where id = ?",
+            (new_first_name, new_last_name, id_to_upd)
```
Sofbog, line 36: I don't think `id_to_upd` is a very good abbreviation. It only saves you two characters anyway. So why don't you rename the variable into` id_to_update`

Sofbog (changes requested):
LGTM

PR #492, opened 2023-09-07
Solution

Kurta97:


app/managers.py
```diff
@@ -0,0 +1,47 @@
...
+        self._connection.execute(
+            f"INSERT INTO {self._table_name} ("
+            f"first_name, last_name) VALUES (?,?)",
+            (first_name, last_name),
+        )
+        self._connection.commit()
+
+    def all(self) -> List[Actor]:
+        actor_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name FROM {self._table_name}"
```
Sofbog, line 22: When you want tot get all values from table it’s better to use `*`

Sofbog (changes requested):

PR #500, opened 2023-09-16
Solution

mwellick:


app/managers.py
```diff
@@ -0,0 +1,44 @@
...
+    def create(self, first_name: str, last_name: str) -> None:
+        self.connection.execute(
+            f"INSERT INTO {self.table_name}  (first_name, last_name)"
+            f"VALUES (?, ?)",
+            (first_name, last_name)
+
+        )
+        self.connection.commit()
+
+    def all(self) -> list:
```
Bohdan-Salamakha, line 20: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]

Bohdan-Salamakha (changes requested):

app/managers.py
```diff
@@ -0,0 +1,44 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self.connection = sqlite3.connect("cinema")
```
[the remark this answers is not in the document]
Bohdan-Salamakha: not fixed
u can type smth like 'cinema.sqlite3'

Bohdan-Salamakha (changes requested):

app/managers.py
```diff
@@ -0,0 +1,44 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self.connection = sqlite3.connect("cinema_sqlite3_db")
```
YukihiroSM, line 8: sqlite is file-based datbase, so you should have some file like cinema.sqlite3. Add file extension as previously asked. 

In your code: if I will create a file cinema_sqlite3_db - it would be some empty file which is not database.

YukihiroSM (changes requested):

Bohdan-Salamakha (approved):

PR #506, opened 2023-09-24
"Solution"

MaksymUK:


app/managers.py
```diff
@@ -0,0 +1,40 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema_db")
```
Bohdan-Salamakha, line 8: add an extension to the database

app/managers.py
```diff
@@ -0,0 +1,40 @@
...
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} (first_name, last_name) "
+            " VALUES(?, ?)",
+            (first_name, last_name),
+        )
+
+        self._connection.commit()
+
+    def all(self) -> list:
```
Bohdan-Salamakha, line 20: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]

Bohdan-Salamakha (changes requested):

EdAlekseiev (approved):
Good job

PR #511, opened 2023-09-26
solution but I'm not sure

vsafian:
I'm not sure if I solved the task as I should have. The most problematic theme is tests. In bigger parts of the tests, I added the assert_called_once_with method, which just mocks functions because I added auto-increment to the id in the database.


app/managers.py
```diff
@@ -0,0 +1,46 @@
...
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
+            "(first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list[Actor]:
+        actors_cursor = self._connection.execute(
+            f"SELECT id, first_name, last_name FROM {self.table_name}"
```
vasylhnatiuk, line 24:
```suggestion
            f"SELECT * FROM {self.table_name}"
```

vasylhnatiuk (changes requested):
you can test without your tests, only with this code:

# from models import Actor
# from managers import ActorManager
#
# if __name__ == "__main__":
#     Actor.objects = ActorManager()
#
#     Actor.objects.create(first_name="Emma", last_name="Watson")
#     Actor.objects.create(first_name="Daniel", last_name="Radclife")
#     print(Actor.objects.all())
#     Actor.objects.update(2, "Daniel", "Radcliffe")
#     print(Actor.objects.all())
#     Actor.objects.delete(1)
#     print(Actor.objects.all())

vsmutok (approved):
Looks Good 👍

PR #527, opened 2023-10-11
Solution

endernaga:


app/managers.py
```diff
@@ -0,0 +1,40 @@
...
+            (first_name, last_name, primary_key_)
+        )
+        self._connection.commit()
+
+    def delete(self, primary_key_: int) -> None:
+        self._connection.execute(
+            "DELETE FROM actors "
+            "WHERE id = ?",
+            (primary_key_,)
+        )
```
NikOneZ1, line 40: Сonnection commit is missing for delete operation

NikOneZ1 (changes requested):
You can also use an attribute with the table name to then reuse the same name in all database operations

YukihiroSM (approved):
Looks good

PR #539, opened 2023-10-19
task completed(1)

DaniilHinzburh:


app/managers.py
```diff
@@ -0,0 +1,38 @@
...
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.db3")
+
+    table_name = "actors"
+
+    def all(self) -> list[Actor]:
+        cursor = self._connection.execute(f"SELECT * FROM {self.table_name}")
+        return [Actor(*row) for row in cursor]
+
+    def create(self, first_name_: str, last_name_: str) -> None:
```
ArsenPidhoretskyi, line 15:
```suggestion
    def create(self, first_name: str, last_name: str) -> None:
```
No need to add this _ at the end

ArsenPidhoretskyi (changes requested):

app/managers.py
DaniilHinzburh: No,it`s need:     def create(self, first_name_: str, last_name_: str) -> None:
        self._connection.execute(
            f"INSERT INTO {self.table_name} "
           -> f"(first_name, last_name) VALUES (?, ?)",
            (first_name_, last_name_)

ArsenPidhoretskyi: Why you need _ at the end?
It is redundant it will work without it

app/managers.py
```diff
@@ -0,0 +1,38 @@
...
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
+            f"(first_name, last_name) VALUES (?, ?)",
+            (first_name_, last_name_)
```
ArsenPidhoretskyi, line 19:
```suggestion
    def create(self, first_name: str, last_name: str) -> None:
        self._connection.execute(
            f"INSERT INTO {self.table_name} "
            f"(first_name, last_name) VALUES (?, ?)",
            (first_name, last_name)
```

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #545, opened 2023-10-22
Solusion

Andru37:


app/managers.py
```diff
@@ -0,0 +1,51 @@
...
+
+from models import Actor
+
+
+class ActorManager:
+
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema")
+
+    def all(self) -> list:
```
NikOneZ1, line 11: It would be better to indicate that you return a list of Actor instances, not just list

app/managers.py
```diff
@@ -0,0 +1,51 @@
...
+    def update(
+            self,
+            id_to_update: int,
+            new_first_name: str,
+            new_last_name: str
+    ) -> None:
+        self._connection.execute(
+            "UPDATE actors "
+            "SET first_name = (?), last_name = (?) "
+            "WHERE ID = (?)",
```
NikOneZ1, line 37: You able to get rid of the parentheses
```suggestion
            "UPDATE actors "
            "SET first_name = ?, last_name = ?"
            "WHERE ID = ?",
```

app/managers.py
```diff
@@ -0,0 +1,51 @@
...
+        )
+        self._connection.commit()
+
+    def delete(
+            self,
+            id_to_delete: int
+    ) -> None:
+        self._connection.execute(
+            "DELETE FROM actors "
+            "WHERE id = (?)",
```
NikOneZ1, line 48: You able to get rid of the parentheses
```suggestion
            "WHERE id = ?",
```

NikOneZ1 (changes requested):

app/managers.py
Andru37: Yes . I totally agree with you.

app/managers.py
Andru37: Good

ArsenPidhoretskyi (approved):
Well done!

PR #554, opened 2023-10-29
Solution

Naz-iv:


app/managers.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.db3")
+        self.table_name = "actors"
+
+    def all(self) -> None:
+        actors_data = self._connection.execute(
+            "SELECT * FROM actors"
```
fsocie7y, line 12: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM clause from new line.

app/managers.py
```diff
@@ -0,0 +1,38 @@
...
+            f"UPDATE {self.table_name} "
+            "SET first_name = ?, last_name = ? "
+            "WHERE ID = ?",
+            (first_name, last_name, id_to_update)
+        )
+        self._connection.commit()
+
+    def delete(self, id_to_delete: int) -> None:
+        self._connection.execute(
+            f"DELETE FROM {self.table_name} "
```
fsocie7y, line 35: same about from

fsocie7y (changes requested):
no need to add db to yr PR) remove it)

PR #565, opened 2023-10-30
py-actor-manager Solution

RVChornyy:


skoval7 (approved):

app/managers.py
```diff
@@ -0,0 +1,45 @@
...
+            f"INSERT INTO {self.table_name}"
+            "(first_name, last_name)"
+            "VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
+        actor_cursor = self._connection.execute(
+            f"SELECT * FROM {self.table_name}"
```
fsocie7y, line 22: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM clause from new line.

app/managers.py
```diff
@@ -0,0 +1,45 @@
...
+            f" SET first_name = ?, last_name = ? WHERE id = ?",
+            (name_to_update,
+             surname_to_update,
+             id_to_update)
+        )
+        self._connection.commit()
+
+    def delete(self, id_to_delete: int) -> None:
+        self._connection.execute(
+            f"DELETE FROM {self.table_name}"
```
fsocie7y, line 41: same about from

fsocie7y (changes requested):

PR #567, opened 2023-10-31
Solution

OlehShumov:


sakhaline (approved):

app/managers.py
```diff
@@ -0,0 +1,42 @@
...
+        self._connection.execute(
+            f"INSERT INTO {self.table_name}"
+            f"(first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
+        actors_data = self._connection.execute(
+            "SELECT * FROM actors"
```
fsocie7y, line 20: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM clause from new line.

fsocie7y (changes requested):
Delete DB from your PR

PR #579, opened 2023-10-31
Implement py-actor-manager

SevKrok:


oleksandr-pupchenko (approved):

app/managers.py
```diff
@@ -0,0 +1,59 @@
+import sqlite3
```
fsocie7y, line 1: Delete DB from your PR

fsocie7y (changes requested):

PR #590, opened 2023-11-11
Solution by Dyriv

IvDyriv:


app/models.py
```diff
@@ -0,0 +1,9 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class Actor:
+
```
NikOneZ1, line 6: You don't need this empty line

app/managers.py
```diff
@@ -0,0 +1,44 @@
...
+            f"INSERT INTO {self.table_name} "
+            "(first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
+        actor_cursor = self._connection.execute(
+            "SELECT id, first_name, last_name "
+            f"FROM {self.table_name}"
```
NikOneZ1, line 22: You can select all fields 
```suggestion
            "SELECT * "
            f"FROM {self.table_name}"
```

NikOneZ1 (changes requested):

app/models.py
IvDyriv: Thank you

app/managers.py
IvDyriv: Thank you, i forgot about it)

YukihiroSM (approved):
Looks good to me!

PR #592, opened 2023-11-13
Solution

olgierrd:


app/managers.py
```diff
@@ -0,0 +1,45 @@
+import sqlite3
```
viktoria-rybenchuk, line 1: add a blank line between each group of imports

viktoria-rybenchuk (changes requested):
Make sure you don't push db files (files with .sqlite, .db3, etc. extension). Read checklist

viktoria-rybenchuk (approved):
Good job!

PR #597, opened 2023-11-19
solution

Andrey1104:


app/managers.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connect = sqlite3.connect("cinema")
+        self.table = "actors"
+
+    @staticmethod
+    def check_name(first_name: str, last_name: str) -> list:
+        if isinstance(first_name and last_name, str):
```
Dimosphen1, line 13: This check will not work correctly, separate this check between two conditions. E.g., `isinstance(3 and "a", str)` will return `True`.

Dimosphen1 (changes requested):
Several changes were requested.

app/managers.py
Andrey1104: this is so sad, I thought it would work

app/managers.py
```diff
@@ -0,0 +1,53 @@
...
+            f"INSERT INTO {self.table} "
+            "(first_name, last_name) "
+            "VALUES (?, ?)",
+            (*self.check_name(first_name, last_name),)
+        )
+        self._connect.commit()
+
+    def all(self) -> list:
+        cursor = self._connect.execute(
+            f"SELECT id, first_name, last_name FROM {self.table};"
```
pavlejviki, line 29: you can use * to retrieve all columns from a table

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #650, opened 2024-01-22
Solution

mrve4nick:


app/managers.py
```diff
@@ -0,0 +1,42 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self.conn = sqlite3.connect("../cinema.sqlite")
```
Dimosphen1, line 8: You could provide only the DB name in the connection without a path here.

Dimosphen1 (approved):
LGTM, one comment is left for consideration.

PR #654, opened 2024-01-29
Solution

Acik33:


app/managers.py
```diff
@@ -0,0 +1,47 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect(
+            "/home/arsen/PycharmProjects/MateAcademy"
```
vasylhnatiuk, line 9: don't use full path. you should write only db name

app/managers.py
```diff
@@ -0,0 +1,47 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect(
```
vasylhnatiuk, line 8:
```suggestion
        self._connection = sqlite3.connect("cinema.sqlite")
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #663, opened 2024-02-06
Develop

yablonskyiv:


.flake8
EdAlekseiev, line 1: Rollback these changes, please

EdAlekseiev (changes requested):

EdAlekseiev (changes requested):
Fix flake8. If you have problems with it, you can write to chat or visit QnA

PR #667, opened 2024-02-07
Solution

isikach:


app/managers.py
```diff
@@ -0,0 +1,48 @@
...
+            (first_name, last_name)
+        )
+        self._conn.commit()
+
+    def all(self) -> None:
+        actors_cursor = self._conn.execute(
+            f"SELECT * FROM {self._table_name}"
+        )
+        for row in actors_cursor:
+            print(row)
```
VladKyianytsia, line 27: all - a method that returns a list of Actor instances from DB (read README)

VladKyianytsia (changes requested):

app/managers.py
isikach: Good point, thanks)

PR #673, opened 2024-02-08
Solution

NataliiaKorzh:


app/managers.py
```diff
@@ -0,0 +1,42 @@
...
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("../cinema.sqlite")
+        self.table_name = "actors"
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
+            f"(first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name,)
+        )
```
aLEKS-e3, line 17: why didn't you commit the operation here? and also you should commit in the delete method

aLEKS-e3 (changes requested):

PR #730, opened 2024-03-10
Solution

Bujhm99:


app/managers.py
```diff
@@ -0,0 +1,26 @@
+import sqlite3
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self.connect = sqlite3.connect("../cinema.sqlite")
+
+    def create(self, first_name: str, last_name: str) -> None:
+        sql = "Insert Into actors (first_name,last_name) VALUES (?,?)"
```
LiudmylaKulinchenko, line 9: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to use uppercase for commands and split query in lines: (
            f"INSERT INTO {self._table_name} "
            "(first_name, last_name) VALUES (?, ?)"
        )
        
 Same with other queries

LiudmylaKulinchenko: You can also try to rewrite without creating additional variable

app/managers.py
```diff
@@ -0,0 +1,26 @@
+import sqlite3
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self.connect = sqlite3.connect("../cinema.sqlite")
```
LiudmylaKulinchenko, line 6: You can store table name as an attribute and then use it in the sql, to avoid hardcoding

app/managers.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def create(self, first_name: str, last_name: str) -> None:
+        sql = "Insert Into actors (first_name,last_name) VALUES (?,?)"
+        self.connect.execute(sql, (first_name, last_name))
+        self.connect.commit()
+
+    def all(self) -> None:
+        actors = self.connect.execute("Select * From actors")
+        print([actor for actor in actors])
```
LiudmylaKulinchenko, line 15: remove print

app/managers.py
```diff
@@ -0,0 +1,26 @@
...
+        return actors
+
+    def update(self, id: int, first_name: str, last_name: str) -> None:
+        sql = "Update actors Set first_name = ?, last_name = ? Where id = ?"
+        self.connect.execute(sql, (first_name, last_name, id))
+        # self.connect.commit()
```
LiudmylaKulinchenko, line 21: uncomment this line

app/managers.py
Bujhm99: Is it ok to stay sql variable?

app/managers.py
Bujhm99: Why it's updating, but, for compare, not deleting?

cinema.sqlite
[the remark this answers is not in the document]
Bujhm99: removed db from commiting, but how to remove it from this pr?

LiudmylaKulinchenko (changes requested):
Please, try to find a solution by yourself to improve your skills and then ping us in chat if you didn't find a right solution

Bujhm99:
removed from pr

app/managers.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self.connect = sqlite3.connect("../cinema.sqlite")
+        self.table_name = "actors"
+
+    def create(self, first_name: str, last_name: str) -> None:
+        sql = f"INSERT INTO {self.table_name} " \
+              "(first_name,last_name) VALUES (?,?)"
```
Sofbog, line 11: Don't use f-strings for SQL. Use parametrization.
 fix all over the code

Sofbog (changes requested):

app/managers.py
Oleksl888: This hasn't been fixed, please parametrise database name

app/managers.py
```diff
@@ -0,0 +1,28 @@
+import sqlite3
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self.connect = sqlite3.connect("../cinema.sqlite")
+
+    def create(self, first_name: str, last_name: str) -> None:
+        sql = "INSERT INTO actors" \
```
Oleksl888, line 9: Avoid using line breakers. It is a common practice to use triple quotes to write sql queries in python. Fix below as well

app/main.py
```diff
@@ -1,13 +1,12 @@
...
-#     print(Actor.objects.all())
-#     Actor.objects.update(2, "Daniel", "Radcliffe")
-#     print(Actor.objects.all())
-#     Actor.objects.delete(1)
-#     print(Actor.objects.all())
+from models import Actor
+from managers import ActorManager
+
+if __name__ == "__main__":
+    Actor.objects = ActorManager()
```
Oleksl888, line 5: The task here is to test methods that you created. Simply printing output does not imply a valid test. You can write simple test cases: for example you create Actor with name = "Emma Watson" and on next line get Actor with this name from database to check that record was created. Do the same for all 4 methods

Oleksl888 (changes requested):

app/managers.py
```diff
@@ -0,0 +1,29 @@
+import sqlite3
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self.connect = sqlite3.connect("../cinema.sqlite")
+
+    def create(self, first_name: str, last_name: str) -> None:
+        sql = "INSERT INTO actors (first_name,last_name) VALUES (?,?)"
```
Oleksl888, line 9: No need to save a query in a separate varibale here. You can simply:
```
self._conn.execute(
            """
                INSERT INTO actors
                (first_name, last_name) VALUES (?, ?)
                """, (first_name, last_name)
        )
        self._conn.commit()

Oleksl888 (approved):
Approve but please fix the notes by Liudmila

Sofbog (approved):

PR #733, opened 2024-03-13
Done

DenZatolokin:


app/cinema_db.sqlite
EdAlekseiev, line 1: Add *.sqlite fiels to .gitignore and delete it from repository

EdAlekseiev (approved):
Good job

PR #735, opened 2024-03-16
Solution

IjdmLoverI:


app/managers.py
```diff
@@ -0,0 +1,59 @@
...
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
+        actor_manager_cursor = self._connection.execute(
+            "SELECT * FROM actors"
+        )
+        # print(f"actor_manager_cursor: {actor_manager_cursor}")
+        # for row in actor_manager_cursor:
+        #     print(row)
```
vsmutok, line 25: remove this comment

app/managers.py
```diff
@@ -0,0 +1,59 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect(
+            "/MateProjectsPython/py-actor-manager/cinema.sqlite"
```
vsmutok, line 9: using absolute path is bad practice, as another developer will have a different path, 
Use `OS` to specify absolute path

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #749, opened 2024-04-01
Solution

Boyok1337:


Bujhm99 (approved):
LGTM

app/managers.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("../cinema.sqlite")
+        self.table_name = "actors"
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
```
Oleksl888, line 16: Overall inserting non-parametrized data with f-string formatting is a bad practice. Here is no need to do it since you work with only one table and no need to get into the habit of doing this.
Also please follow SQL formatting style. Your statement should be rather broken down like so:
```
VALUES (?, ?).
Also it is a common practice to use triple quotes when writing raw SQL with multilines

File: cinema.sqlite
Comment (Oleksl888) [inline cinema.sqlite:1]: Please remove database

**Review by Oleksl888 — CHANGES_REQUESTED** [review]

**Review by Oleksl888 — CHANGES_REQUESTED** [review]
Flake8 check is failing

File: app/managers.py [hunk 2]
```diff
@@ -0,0 +1,47 @@
...
+        actors_cursor = self._connection.execute(
+            f"SELECT * FROM {self.table_name}"
+        )
+        return [
+            Actor(*row) for row in actors_cursor
+        ]
+
+    def update(self, actor_id: int, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"UPDATE {self.table_name} "
```
Oleksl888, line 34: why not use triple quotes everywhere then?

app/managers.py
```diff
@@ -0,0 +1,47 @@
...
+            INSERT INTO actors(first_name, last_name)
+            VALUES (?, ?)
+            """,
+            (first_name, last_name),
+        )
+        self._connection.commit()
+
+    def all(self) -> List[Actor]:
+        actors_cursor = self._connection.execute(
+            f"SELECT * FROM {self.table_name}"
```
Oleksl888, line 26: If you follow SQL formatting and best practices this line should look like so:
```
f"""
SELECT * 
FROM {self.table_name}
"""
Each sql keyword on new line, quotes are on a different level from actualt query

Oleksl888 (approved):
Approved with notes

PR #772, opened 2024-04-15
Solution for task

oleksandra-shershen:


cinema.sqlite
Oleksl888, line 1: Please remove database

app/managers.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("../cinema.sqlite")
+        self.table_name = "actors"
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
```
Oleksl888, line 13: Overall inserting non-parametrized data with f-string formatting is a bad practice. Here is no need to do it since you work with only one table and no need to get into the habit of doing this.
Also please follow SQL formatting style. Your statement should be rather broken down like so:
```
VALUES (?, ?)

File: app/managers.py [hunk 2]
```diff
@@ -0,0 +1,44 @@
...
+        return [Actor(*row) for row in actors_cursor]
+
+    def update(
+            self,
+            id_to_update: int,
+            new_first_name: str,
+            new_last_name: str
+    ) -> None:
+        self._connection.execute(
+            f"UPDATE {self.table_name} "
```
Oleksl888, line 32: it ss a common practice to use triple quotes when writing raw SQL with multilines

app/managers.py
```diff
@@ -0,0 +1,44 @@
...
+            f"SET first_name = ?, last_name = ? "
+            f"WHERE id = ?",
+            (new_first_name, new_last_name, id_to_update)
+        )
+        self._connection.commit()
+
+    def delete(self, id_to_delete: int) -> None:
+        self._connection.execute(
+            f"DELETE FROM {self.table_name} WHERE id = ?",
```
Oleksl888, line 41: Even though it fits in one line by sql standards is better to break statements by main keywords:
```
DELETE
FROM {self.table}
WHERE id = ?

Oleksl888 (changes requested):

app/managers.py
oleksandra-shershen: > Overall inserting non-parametrized data with f-string formatting is a bad practice. Here is no need to do it since you work with only one table and no need to get into the habit of doing this. Also please follow SQL formatting style. Your statement should be rather broken down like so:
> 
> ```
> INSERT INTO actors (first_name, last_name) 
> VALUES (?, ?)
> ```

Resolved

app/managers.py
oleksandra-shershen: Resolved

fsocie7y (approved):
GJ 👍

PR #774, opened 2024-04-16
Solution

Nyxan:


app/managers.py
```diff
@@ -0,0 +1,44 @@
...
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} (first_name, last_name) "
+            f"VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list[Actor]:
+        actors_cursor = self._connection.execute(
+            f"SELECT * FROM {self.table_name}"
```
fsocie7y, line 23: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM clause from new line.

app/managers.py
```diff
@@ -0,0 +1,44 @@
...
+        self._connection.execute(
+            f"UPDATE {self.table_name} "
+            f"SET first_name = ?, last_name = ? WHERE id = ?",
+            (first_name, last_name, id_to_update)
+        )
+        self._connection.commit()
+
+    def delete(self, id_to_delete: int) -> None:
+        self._connection.execute(
+            f"DELETE FROM {self.table_name} WHERE id = ?",
```
fsocie7y, line 42: the same

fsocie7y (changes requested):

app/managers.py
```diff
@@ -0,0 +1,48 @@
...
+        return [Actor(*row) for row in actors_cursor]
+
+    def update(
+            self,
+            id_to_update: int,
+            first_name: str,
+            last_name: str
+    ) -> None:
+        self._connection.execute(
+            f"UPDATE {self.table_name} "
```
Oleksl888, line 35: you can use triple quotes for raw sql queries - it is a common practice

app/managers.py
```diff
@@ -0,0 +1,48 @@
+from __future__ import annotations
```
Oleksl888, line 1: do you really need it here?

app/managers.py
```diff
@@ -0,0 +1,48 @@
...
+
+import sqlite3
```
Oleksl888, line 3: must be blank line between diffrent import types

Oleksl888 (approved):
Approved! But please check small details to fix

PR #780, opened 2024-04-16
Solution

DmytroHlazyrin:


app/managers.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("../cinema.sqlite")
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            """
+            INSERT INTO actors
```
Oleksl888, line 13: here is better to make a new line for a new keyword
```
VALUES (?,?)

Oleksl888 (approved):
Good job!

PR #785, opened 2024-04-17
solution

alonasorochynska:


app/managers.py
```diff
@@ -0,0 +1,43 @@
...
+        self._connection.execute(
+            """INSERT INTO actors (first_name, last_name)
+            VALUES (?, ?)""",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list[Actor]:
+        actors_cursor = self._connection.execute(
+            """SELECT * FROM actors"""
```
Oleksl888, line 20: even if it fits in one line, sql standard of formatting suggests to start key words from new line. Also is common to put triple quotes on a separate line like so:
```
"""
SELECT * 
FROM actors
"""

 Please check in other queries as well if formatiing style is in compliance

Oleksl888 (changes requested):

fsocie7y (approved):
GJ 👍

PR #786, opened 2024-04-17
Solution

y-kondrashova:


app/managers.py
```diff
@@ -0,0 +1,50 @@
...
+            WHERE id = ?
+            """,
+            (id_to_delete,)
+        )
+        self.connection.commit()
+
+    def all(self) -> list[Actor]:
+        cursor = self.connection.execute(
+            """
+            SELECT * FROM actors
```
Oleksl888, line 47: Here you could also follow sql formatting:
```
SELECT *
FROM actors

Oleksl888 (approved):
Good job!

PR #797, opened 2024-04-19
Solution

siaserb:


app/managers.py
```diff
@@ -0,0 +1,47 @@
...
+        self._connection.execute(
+            f"INSERT INTO {self._table_name} (first_name, last_name) "
+            f"VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list[Actor]:
+        cursor = self._connection.execute(
+            f"SELECT * FROM {self._table_name}"
```
Oleksl888, line 21: as per sql style guidelines must be a new line for each keyword

app/managers.py
```diff
@@ -0,0 +1,47 @@
...
+            f"UPDATE {self._table_name} "
+            f"SET first_name = ?, last_name = ? "
+            f"WHERE id = ?",
+            (first_name, last_name, id_to_update)
+        )
+        self._connection.commit()
+
+    def delete(self, id_to_delete: int) -> None:
+        self._connection.execute(
+            f"DELETE FROM {self._table_name} WHERE id = ?",
```
Oleksl888, line 44: fix formatting here also

app/managers.py
```diff
@@ -0,0 +1,47 @@
...
+        ]
+
+    def update(
+            self,
+            id_to_update: int,
+            first_name: str,
+            last_name: str
+    ) -> None:
+        self._connection.execute(
+            f"UPDATE {self._table_name} "
```
Oleksl888, line 35: it is a common practice to use triple quotes for multiline sql queries

Oleksl888 (changes requested):

app/managers.py
Oleksl888: https://www.sqlstyle.guide/
check styling here for examples

YuriiKindrat (changes requested):
remove db from ur PR

PR #801, opened 2024-04-21
Solution

olga-kim7:


app/manager.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name}"
+            f" (first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name,)
+        )
+        self._connection.commit()
+
+    def all(self) -> None:
```
Dimosphen1, line 19: Correct your return type annotation. Indicate instances of what class you return in a list in the return type annotation

Dimosphen1 (changes requested):
Several changes were requested

app/manager.py
olga-kim7: Done

Polyakiv-Andrey (approved):
Good job!

PR #806, opened 2024-04-23
Solution

VladyslavKsh:


app/managers.py
```diff
@@ -0,0 +1,45 @@
...
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} (first_name, last_name) "
+            "VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list[Actor]:
+        cinema_cursor = self._connection.execute(
+            f"SELECT * FROM {self.table_name}"
```
Oleksl888, line 21: follow SQL formatting guidelines

app/managers.py
```diff
@@ -0,0 +1,45 @@
...
+        return [
+            Actor(*row) for row in cinema_cursor
+        ]
+
+    def update(self,
+               id_to_update: int,
+               first_name: str,
+               last_name: str) -> None:
+        self._connection.execute(
+            f"UPDATE {self.table_name} "
```
Oleksl888, line 32: it is common to use triple quotes when writing multiline sql

Oleksl888 (changes requested):

app/managers.py
Oleksl888: Correct would be:
```
SELECT *
FROM table_name

File: app/managers.py [hunk 3]
```diff
@@ -0,0 +1,43 @@
...
+            f"""UPDATE {self.table_name}
+            SET first_name = ?, last_name = ?
+            WHERE id = ?""",
+            (first_name, last_name, id_to_update)
+        )
+        self._connection.commit()
+
+    def delete(self, id_to_delete: int) -> None:
+        self._connection.execute(
+            f"""DELETE FROM {self.table_name}
```
Oleksl888, line 39: Correct would be:
```
DELETE
FROM table_name
WHERE id = ?

Oleksl888 (approved):
Approved. But read more about sql styling here: https://www.sqlstyle.guide/

PR #816, opened 2024-05-02
solution

RomanDanyl:


app/managers.py
```diff
@@ -0,0 +1,40 @@
+import sqlite3
+from models import Actor
+
+
+class ActorManager:
+
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect(
+            "C:/progects/py-actor-manager/app/cinema_file.sqlite"
```
pavlopro, line 9: There should not be an explicit path to a specific database.
Instead, just set the desired database name. For example, cinema_file.sqlite.

pavlopro (changes requested):
From checklist:
Don't Push db files

Polyakiv-Andrey (approved):
Good job!

PR #823, opened 2024-05-13
Solution

sind14:


vasylhnatiuk (changes requested):
delete cinema.sqlite

vasylhnatiuk (changes requested):
you should also add it in git ignore because you added again

vasylhnatiuk (changes requested):
and delete cinema.sqlite again

vasylhnatiuk (approved):

PR #834, opened 2024-05-29
'Solution'

Julia15031988:


managers.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def create(self, format_: str):
+        self._connection.execute(
+            f" INSERT INTO {self.table_name} (format) VALUES (?)",
+            (format_,)
+
+        )
+        self._connection.commit()
+
+    def all(self):
```
vasylhnatiuk, line 19: add return annotation here and in all of methods

vasylhnatiuk (changes requested):

managers.py
```diff
@@ -0,0 +1,43 @@
+import sqlite3
+from models import Actor
+from typing import List
+
+class ActorManager:
+    def __init__(self):
```
vasylhnatiuk, line 6:
```suggestion
    def __init__(self) -> None:

```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #835, opened 2024-05-30
'Solution'

sdannyk:


app/managers.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} (first_name, last_name)"
+            f" VALUES (?, ?)",
+            (first_name, last_name,)
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
```
vasylhnatiuk, line 20:
```suggestion
    def all(self) -> list[Actor]:

```

vasylhnatiuk (changes requested):

app/cinema.sqlite
EdAlekseiev, line 1: You should add db file to .gitignore and delete from GitHub, because this file can be large

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #848, opened 2024-06-14
solution

AndrewwSSS:


app/main.py
```diff
@@ -1,13 +1,13 @@
...
+if __name__ == "__main__":
+    Actor.objects = ActorManager()
+
+    Actor.objects.create(first_name="Emma", last_name="Watson")
+    Actor.objects.create(first_name="Daniel", last_name="Radclife")
+    print(Actor.objects.all())
+    Actor.objects.update(2, "Daniel", "Radcliffe")
+    print(Actor.objects.all())
+    Actor.objects.delete(1)
+    print(Actor.objects.all())
```
DevSpaciX, line 13: All this prints are necessery ? Maybe better delete them ?)

cinema.sqlite
DevSpaciX, line 1: Delete db from commit file : ) It should not be here

DevSpaciX (changes requested):

app/managers.py
```diff
@@ -0,0 +1,40 @@
...
+            Actor(*row) for row in cursor
+        ]
+
+    def update(self, id: int,
+               first_name: str,
+               second_name: str) -> None:
+        self.connection.execute(
+            f"UPDATE {self.table_name} "
+            f"SET first_name=?, second_name=? WHERE id=?",
+            (first_name, second_name, id)
```
DevSpaciX, line 32: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM clause from new line.

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #853, opened 2024-06-17
Solution_v1

BileichukIvan:


app/managers.py
```diff
@@ -0,0 +1,42 @@
...
+            f"WHERE id = ?",
+            (new_first_name, new_last_name, id_to_update),
+        )
+        self._connection.commit()
+
+    def delete(self, id_to_delete: int) -> None:
+        self._connection.execute(
+            f"DELETE FROM {self.table_name} WHERE id = ?",
+            (id_to_delete, ),
+        )
```
DevSpaciX, line 41: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM clause from new line.

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #854, opened 2024-06-18
Solution

SrMusienko:


app/managers.py
```diff
@@ -0,0 +1,44 @@
...
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} (first_name, last_name) "
+            f"VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
+        actor_cursor = self._connection.execute(
```
DevSpaciX, line 19: `-> list:`

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #855, opened 2024-06-18
sol1

Unlie9:


managers.py
```diff
@@ -0,0 +1,53 @@
...
+        self._connection.commit()
+
+    def all(self):
+        actors_cursor = self._connection.execute(
+            f"SELECT * FROM {self.table_name}"
+        )
+
+        return [
+            Actor(*row) for row in actors_cursor
+        ]
```
DevSpaciX, line 25: Add return type hint everywhere : )

managers.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def update(self, id_to_update: int, new_first_name: str, new_last_name: str):
+        if new_first_name and not new_last_name:
+            self._connection.execute(
+                f"UPDATE {self.table_name} SET first_name = ? WHERE id = ? ",
+                (new_first_name, id_to_update)
```
DevSpaciX, line 31: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM clause from new line.

DevSpaciX (changes requested):

managers.py
```diff
@@ -0,0 +1,53 @@
...
+        actors_cursor = self._connection.execute(
+            f"SELECT * FROM {self.table_name}"
+        )
+
+        return [
+            Actor(*row) for row in actors_cursor
+        ]
+
+    def update(self, id_to_update: int, new_first_name: str, new_last_name: str):
+        if new_first_name and not new_last_name:
```
DevSpaciX, line 28: it is not id_to_update ; It is actor id

DevSpaciX (approved):

PR #857, opened 2024-06-18
Solution

Yuruch:


app/managers.py
```diff
@@ -0,0 +1,47 @@
...
+        self._connection.execute(
+            f"UPDATE {self.table} "
+            f"SET first_name = ?, last_name = ? "
+            f"WHERE id = ?",
+            (first_name, last_name, id_to_update)
+        )
+        self._connection.commit()
+
+    def delete(self, id_to_delete: str) -> None:
+        self._connection.execute(
```
DevSpaciX, line 42: It is not id_to_delete . It is actor_id : )

DevSpaciX (changes requested):

app/managers.py
```diff
@@ -0,0 +1,47 @@
...
+        cursor = self._connection.execute(
+            f"SELECT * FROM {self.table}"
+        )
+        return [
+            Actor(*row) for row in cursor
+        ]
+
+    def update(
+            self,
+            id_to_update: str,
```
DevSpaciX, line 29: same here , why it is called id_to_update if it actually an actor id ?)

DevSpaciX (changes requested):

app/managers.py
Yuruch: we had name like this in the video, so i thought this is a good name 
image

DevSpaciX: Imho that was more for example : )

DevSpaciX (approved):

PR #862, opened 2024-06-18
add solution

jpdreamthug:


app/managers.py
```diff
@@ -0,0 +1,47 @@
...
+               last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} (first_name, last_name) "
+            f"VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
+        cinema_actors_cursor = self._connection.execute(
```
DevSpaciX, line 22: List of what ?) Maybe Actor ?)

app/managers.py
```diff
@@ -0,0 +1,47 @@
...
+            f"SELECT * FROM {self.table_name}"
+        )
+        return [
+            Actor(*row) for row in cinema_actors_cursor
+        ]
+
+    def update(self,
+               id_to_update: int,
+               new_first_name: str,
+               new_last_name: str) -> None:
```
DevSpaciX, line 32: It is not id to update . It is actor id

DevSpaciX (changes requested):

app/managers.py
```diff
@@ -0,0 +1,47 @@
...
+               last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} (first_name, last_name) "
+            f"VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
+        actors = self._connection.execute(
```
DevSpaciX, line 22: `def all(self) -> list[Actor]:` will be more informative , is not it ?

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #863, opened 2024-06-18
Solution V.1

SemenetsSergii:


app/managers.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table} (first_name, last_name) "
+            f"VALUES (?, ?) ",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> None:
```
DevSpaciX, line 18: Why `-> None:` ?? Maybe `list` ?

DevSpaciX (changes requested):

app/managers.py
```diff
@@ -0,0 +1,42 @@
...
+    ) -> None:
+        self._connection.execute(
+            f"UPDATE {self.table} "
+            "SET first_name = ? AND last_name = ? "
+            "WHERE id = ? ",
+            (new_first_name, new_last_name, id_to_update)
+        )
+        self._connection.commit()
+
+    def delete(self, id_to_delete: int) -> None:
```
DevSpaciX, line 37: Also it is not id_to_delete , it is actor id (

app/managers.py
```diff
@@ -0,0 +1,42 @@
...
+        self._connection.execute(
+            f"DELETE FROM {self.table} WHERE id = ? ",
+            (id_to_delete,)
```
DevSpaciX, line 40: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM clause from new line.

DevSpaciX (changes requested):
Forgotten fixes (

app/managers.py
SemenetsSergii: yes, better when other dev can understand :)

DevSpaciX (approved):
Good !

PR #864, opened 2024-06-19
Solution

VShveda:


app/managers.py
```diff
@@ -0,0 +1,44 @@
...
+               last_name: str) -> None:
+        self._connection.execute(
+            f"UPDATE {self.table_name} "
+            f"SET first_name=?, last_name=? "
+            f"WHERE id=?",
+            (id, first_name, last_name)
+        )
+        self._connection.commit()
+
+    def delete(self, id_del: int) -> None:
```
DevSpaciX, line 38: `id_del` ? It can be simple named `id`

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #870, opened 2024-06-19
solution

DimaYankovskiy:


app/managers.py
```diff
@@ -0,0 +1,39 @@
...
+
+from app.models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema_db.sqlite")
+        self.table_name = "actors"
+
+    def all(self) -> list:
```
DevSpaciX, line 11: `-> list:`

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #871, opened 2024-06-19
Solution

Andrii1000:


app/managers.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self.connection.execute(
+            f"INSERT INTO {self.table_name} (first_name, last_name) "
+            f"VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self.connection.commit()
+
+    def all(self) -> list:
```
DevSpaciX, line 19: `-> list:`

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #872, opened 2024-06-19
sol1

IvanVahanov:


managers.py
```diff
@@ -0,0 +1,26 @@
...
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self, db_name: str):
+        self.connection = sqlite3.connect(db_name)
+        self.cursor = self.connection.cursor()
+
+    def create(self, first_name: str, last_name: str):
+        self.cursor.execute("INSERT INTO actors (first_name, last_name) VALUES (?, ?)", (first_name, last_name))
```
DevSpaciX, line 11: `INSERT INTO actors`
Store table name in a viriable . And then just use it .

managers.py
```diff
@@ -0,0 +1,26 @@
...
+        self.connection.commit()
+        return self.cursor.lastrowid
+
+    def all(self):
+        self.cursor.execute("SELECT id, first_name, last_name FROM actors")
```
DevSpaciX, line 16: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM clause from new line.

DevSpaciX, line 16: Add return type hints everywhere : )

managers.py
```diff
@@ -0,0 +1,26 @@
...
+        rows = self.cursor.fetchall()
+        return [Actor(id=row[0], first_name=row[1], last_name=row[2]) for row in rows]
+
+    def update(self, actor_id: int, first_name: str, last_name: str):
+        self.cursor.execute("UPDATE actors SET first_name = ?, last_name = ? WHERE id = ?", (first_name, last_name, actor_id))
+        self.connection.commit()
```
DevSpaciX, line 22: Same ) 
Follow SQL style guidelines

DevSpaciX (changes requested):
Some fixes required

managers.py
```diff
@@ -0,0 +1,26 @@
+import sqlite3
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self, db_name: str):
+        self.connection = sqlite3.connect(db_name)
+        self.cursor = self.connection.cursor()
```
DevSpaciX, line 8: You do not need cursor here : )

PR #873, opened 2024-06-19
'Solution'

Irina17191:


managers.py
```diff
@@ -0,0 +1,40 @@
...
+        self.table_name = "actors"
+
+
+    def all(self):
+        cursor = self._connection.execute(
+            "SELECT * FROM actors"
+        )
+        rows = cursor.fetchall()
+        actors = [Actor(*row) for row in rows]
+        return actors
```
DevSpaciX, line 16: Add retrun type hint everywhere : )

managers.py
```diff
@@ -0,0 +1,40 @@
...
+            f"INSERT INTO {self.table_name} (first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+
+    def update(self, id_to_update: int, new_first_name: str, new_last_name: str):
+        self._connection.execute(
+            f"UPDATE {self.table_name} SET first_name = ?, last_name = ? WHERE id = ?",
+            (new_first_name, new_last_name, id_to_update)
```
DevSpaciX, line 30: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM clause from new line.
Fix it everywhere : )

managers.py
```diff
@@ -0,0 +1,40 @@
...
+
+class ActorManager:
+    def __init__(self):
+        self._connection = sqlite3.connect("cinema.sqlite")
+        self.table_name = "actors"
+
+
+    def all(self):
+        cursor = self._connection.execute(
+            "SELECT * FROM actors"
```
DevSpaciX, line 12: Why do you using hardcoded table name here "SELECT * FROM actors"  instead of using your viriable as you did in other places ?)

DevSpaciX (changes requested):

managers.py
```diff
@@ -0,0 +1,40 @@
...
+    def create(self, first_name: str, last_name: str):
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} (first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+
+    def update(self, id_to_update: int, new_first_name: str, new_last_name: str):
+        self._connection.execute(
```
DevSpaciX, line 28: Actually it is not id_to_update . It is actor id : )

DevSpaciX (approved):

PR #874, opened 2024-06-19
'Solution'

ghost:


app/managers.py
```diff
@@ -0,0 +1,43 @@
...
+            f"SET {actor_name} = ? "
+            f"WHERE id = ? ",
+            (actor_surname, actor_id,)
+        )
+        self._connection.commit()
+
+    def delete(self, actor_id: int) -> None:
+        self._connection.execute(
+            f"DELETE FROM {self.table_name} WHERE id = ? ",
+            (actor_id,)
```
DevSpaciX, line 41: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM clause from new line.

DevSpaciX (changes requested):

app/managers.py
```diff
@@ -0,0 +1,48 @@
...
+        self._connection.execute(
+            f"DELETE "
+            f"FROM {self.table_name} WHERE id = ? ",
+            (actor_id,)
+        )
+        self._connection.commit()
+
+
+manager = ActorManager()
+manager.update(1, "Asd", "123")
```
DevSpaciX, line 48: What is it for ?)

app/managers.py
```diff
@@ -0,0 +1,48 @@
...
+            f"WHERE id = ? ",
+            (actor_surname, actor_id,)
+        )
+        self._connection.commit()
+
+    def delete(self, actor_id: int) -> None:
+        self._connection.execute(
+            f"DELETE "
+            f"FROM {self.table_name} WHERE id = ? ",
+            (actor_id,)
```
DevSpaciX, line 42: WHERE stmt should be started from the new line ) I already told about it in the previous comment

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #875, opened 2024-06-20
"Solution"

MaglorFeanorson:


app/managers.py
```diff
@@ -0,0 +1,50 @@
...
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
+        actor_cursor = self._connection.execute(
+            f"SELECT * FROM {self.table}"
+        )
+        return [
+            Actor(*row) for row in actor_cursor
+        ]
```
DevSpaciX, line 23: `-> list:`

app/managers.py
```diff
@@ -0,0 +1,50 @@
...
+            f"DELETE FROM {self.table} WHERE id = ? ",
+            (id_to_delete,)
+        )
+        self._connection.commit()
+
+
+if __name__ == "__main__":
+    manager = ActorManager()
+    manager.delete(15)
+    print(manager.all())
```
DevSpaciX, line 50: Is this code for test ?

app/managers.py
```diff
@@ -0,0 +1,50 @@
...
+            "WHERE id = ?",
+            (new_first_name, new_last_name, id_to_update)
+        )
+        self._connection.commit()
+
+    def delete(self, id_to_delete: int) -> None:
+        self._connection.execute(
+            f"DELETE FROM {self.table} WHERE id = ? ",
+            (id_to_delete,)
+        )
```
DevSpaciX, line 43: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM clause from new line.

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #884, opened 2024-06-28
py actor manager solution

TenbReala:


app/managers.py
```diff
@@ -0,0 +1,38 @@
...
+            f"(first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name)
+        )
+
+    def all(self) -> list[Actor]:
+        actors_cursor = self.connection.execute(
+            f"SELECT * FROM {self.table_name}"
+        )
+        return [
+            Actor(id=actor_id, first_name=first_name, last_name=last_name)
```
pavlopro, line 23: Simlify this to 
`return [Actor(*row) for row in actors_cursor]
`as it was in your previous commit

pavlopro (changes requested):

PR #905, opened 2024-07-17
D_ORM-actor-manager

DOLC3-GABBANA:


app/managers.py
```diff
@@ -0,0 +1,50 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("db/Cinema.db")
```
Dimosphen1, line 8: Use `os.path` in the future to define paths that would work independently from directories delimiters on different operating systems

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

PR #907, opened 2024-07-19
Solution

Wallian169:


managers.py
```diff
@@ -0,0 +1,52 @@
...
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
+        actor_manager_cursor = self._connection.execute(
+            f"SELECT * FROM {self._table_name};"
+        )
+        return [
+            Actor(*actor) for actor in actor_manager_cursor.fetchall()
```
pavlopro, line 23: You can omit fetchall use

managers.py
```diff
@@ -0,0 +1,52 @@
...
+        self._connection.execute(
+            f"DELETE FROM {self._table_name} "
+        )
+        self._connection.commit()
+
+    def reset(self) -> None:
+        self._connection.execute(
+            f"VACUUM"
+        )
+        self._connection.commit()
```
pavlopro, line 52: No need to have those last two methods. Please remove them

pavlopro (changes requested):
Please remove db file from PR

pavlopro (changes requested):
Good Job!

pavlopro (approved):
Good Job!

PR #943, opened 2024-08-20
implement actor_manager

vladislav-tsybuliak1:


.gitignore
```diff
@@ -7,3 +7,4 @@ venv/
 .pytest_cache/
 **__pycache__/
 **db.sqlite3
+*.sqlite
```
Oleksl888, line 10: no need to push changes to this file

app/main.py
```diff
@@ -1,13 +1,14 @@
...
-#     Actor.objects.update(2, "Daniel", "Radcliffe")
-#     print(Actor.objects.all())
-#     Actor.objects.delete(1)
-#     print(Actor.objects.all())
+from app.models import Actor
+from app.managers import ActorManager
+
+
+if __name__ == "__main__":
+    Actor.objects = ActorManager()
```
Oleksl888, line 6: no need to push changes to this file

app/managers.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class ActorManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.sqlite")
+        self.table_name = "actors"
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
```
Oleksl888, line 13: It is common to use triple quotes for multilne queries

Oleksl888 (changes requested):

app/managers.py
vladislav-tsybuliak1: Ok, but then there is a need to change the videos on the platform, because the lecturer there doesn't do that...

oliasymak24 (approved):

PR #951, opened 2024-08-21
Solution

mihavryliuk:


app/managers.py
```diff
@@ -0,0 +1,48 @@
...
+        self._connection.execute(
+            f"""
+        INSERT INTO {self.table_name} (first_name, last_name)
+        VALUES (?, ?)
+        """,
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
```
Oleksl888, line 21: you can be more speicific about type in list contents

app/managers.py
```diff
@@ -0,0 +1,48 @@
...
+        ]
+
+    def update(
+            self,
+            id_to_update: int,
+            new_first_name: str,
+            new_last_name: str
+    ) -> None:
+        self._connection.execute(
+            f"UPDATE {self.table_name} "
```
Oleksl888, line 36: would make sense to use triple quotes here as well

cinema.sqlite
Oleksl888, line 1: no need to push db

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #959, opened 2024-08-22
implement py-actor-manager

Maksym-Turenko:


app/managers.py
```diff
@@ -0,0 +1,48 @@
...
+from models import Actor
+
+
+class CinemaManager:
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema.sqlite")
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            "INSERT INTO Actors (first_name, last_name) VALUES (?, ?)",
```
oliasymak24, line 11: You add a table name like an attribute and the pass it in each command

app/managers.py
```diff
@@ -0,0 +1,48 @@
...
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
```
oliasymak24, line 16: This function will return list of what? Fix annotation

app/managers.py
```diff
@@ -0,0 +1,48 @@
...
+        cinema_cursor = self._connection.execute(
+            "SELECT * FROM Actors"
+        )
+        actors = [Actor(*row) for row in cinema_cursor]
+        cinema_cursor.close()
+        return actors
+
+    def update(self, _id: int, new_name: str, new_last_name: str) -> None:
+        self._connection.execute(
+            "UPDATE Actors SET first_name = ?, last_name = ? WHERE id = ?",
```
oliasymak24, line 26: Better to put each command to next line
Also then use triple quotes

app/managers.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+if __name__ == "__main__":
+    manager = CinemaManager()
+    manager.create("Taras", "Bulba")
+    print(manager.all())
+    manager.update(1, "Petya", "Poroh")
+    print(manager.all())
+    manager.delete(1)
+    manager.disconnect()
```
oliasymak24, line 48: Delete here this test code

oliasymak24 (changes requested):

app/managers.py
```diff
@@ -0,0 +1,45 @@
...
+        )
+        actors = [Actor(*row) for row in cinema_cursor]
+        cinema_cursor.close()
+        return actors
+
+    def update(self, _id: int, new_name: str, new_last_name: str) -> None:
+        self._connection.execute(
+            f"UPDATE {self.table_name} "
+            f"SET first_name = ?, last_name = ? "
+            f"WHERE id = ?",
```
oliasymak24, line 31: Use triple quotes, check everywhere

oliasymak24 (changes requested):

Oleksl888 (approved):
Approved but check last comment please

app/managers.py
```diff
@@ -0,0 +1,52 @@
+import sqlite3
+from models import Actor
```
Oleksl888, line 2: should be a blank line between imports + we import like from app.models...

PR #967, opened 2024-08-26
Solution

NazarSerdiuk1:


app/managers.py
```diff
@@ -0,0 +1,48 @@
...
+    ) -> None:
+        self._connection.execute(
+            f"UPDATE {self.table_name} "
+            f"SET first_name = ?, last_name = ? "
+            f"WHERE id = ?",
+            (first_name, last_name, id_to_update),
+        )
+        self._connection.commit()
+
+    def delete(self, id_to_delete: int) -> None:
```
Dimosphen1, line 42: The naming of `id_to_delete` is too explicit considering that the method name itself is `delete`. You could use `actor_id` instead in the future

Dimosphen1 (approved):
LGTM, one comment is left for consideration

PR #972, opened 2024-09-04
'Solution'

AlexTarasov57:


app/managers.py
```diff
@@ -0,0 +1,52 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect(
+            "C:/Users/taras/projects/py-actor-manager/cinema.sqlite"
```
pavlopro, line 10: You should not use your own path
Correct to `cinema.sqlite`

app/managers.py
```diff
@@ -0,0 +1,52 @@
...
+            first_name: str,
+            last_name: str
+    ) -> None:
+        self._connection.execute(
+            f"UPDATE {self.table_name} "
+            f"SET first_name=?, last_name=? "
+            f"WHERE id=?",
+            (first_name, last_name, id_to_update)
+        )
+
```
pavlopro, line 43: Remove this redundant blank line

pavlopro (changes requested):

app/managers.py
AlexTarasov57: when I use the path cinema.sqlite, I see an error (self._connection.execute(
sqlite3.OperationalError: no such table: actors), how can I fix this?

PR #1010, opened 2024-10-23
solution

dmytrik:


app/managers.py
```diff
@@ -0,0 +1,52 @@
...
+        )
+        self._db.commit()
+
+    def delete(self, id: int) -> None:
+        self._db.execute(
+            f"DELETE FROM {self._table_name} "
+            "WHERE id = ? ",
+            (id,)
+        )
+        self._db.commit()
```
Arsen-hrynevych, line 52: Adding a method to close the connection when it's no longer needed is a great idea. Not closing the connection can lead to resource leaks, especially if the application is long-running or frequently opens new connections.

Arsen-hrynevych (approved):
ok

PR #1018, opened 2024-10-30
solution

OlhaStadnik:


skyfoxwork (approved):
Hi :) Good job!)

I think the cinema.sqlite and cinema_db.sqlite files are not needed.

PR #1026, opened 2024-10-30
Solution

Le0n-K:


app/managers.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
+            f"(first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
```
molodsh1y1, line 18: You can use list[Actor] instead of just list. It provides better type clarity in your code.

molodsh1y1 (approved):
LGTM!

PR #1041, opened 2024-11-02
Solution

sumseer:


FursykIvan (approved):
The solution seems normal, but where is the file main.py?

PR #1099, opened 2024-12-30
Solution

julia4406:


app/managers.py
```diff
@@ -0,0 +1,70 @@
...
+            f"VALUES (?, ?)",
+            (first_name, last_name,)
+        )
+        self.connection.commit()
+
+    def all(self) -> list[Actor]:
+        actors_data_cursor = self.connection.execute(
+            f"SELECT * from {self.table_name}"
+        )
+        return [Actor(*row) for row in actors_data_cursor]
```
petrykivd, line 38: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM clause from new line.
Also fix it in all places

petrykivd (changes requested):

app/managers.py
julia4406: Hello. I tried to make this query in separate lines, but Flake8 didn't pass (it found multiple spaces, but they not). May be cause the f-string, because query with hard-coded table-name is good for Flake8.
sql_query = """
SELECT *
FROM actors
"""
I tried to reformat code with black, but Flake8 didn't pass also. 
So, I found only two ways to pass Flake8 - or hard-code table_name, or use one-line query.

Btw, w3schools suggest one-line query in case for selecting all (https://www.w3schools.com/mysql/mysql_select.asp)

petrykivd: def all(self) -> list[Actor]:
        actors_cursor = self._connection.execute(
            f"SELECT * "
            f"FROM {self.table_name}"
        )
        return [Actor(*row) for row in actors_cursor]

petrykivd: It's must work good

petrykivd (approved):
Nice! 😎 I recommend trying to fix the raw SQL 👌

app/managers.py
julia4406: Yeah, it works! 
I had overthinking with this query and didn't use the simplest one )))

PR #1102, opened 2025-01-03
'Solution'

dottfmar:


Dogober (changes requested):
You can see this information into checklist.md

PR #1108, opened 2025-01-04
'Solution'

VovaGld:


app/managers.py
```diff
@@ -0,0 +1,41 @@
...
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> List[Actor]:
+        cursor = self._connection.execute(
+            f"SELECT * FROM {self.table_name}"
+        )
+
+        return list(cursor)
```
DanilTrukhanov, line 26: You are returning here just a cursor as an iterator, instead of a list of the Actor objects

DanilTrukhanov (approved):

app/managers.py
```diff
@@ -0,0 +1,41 @@
...
+            (first_name, last_name, id_)
+        )
+        self._connection.commit()
+
+    def delete(self, id_: int) -> None:
+        self._connection.execute(
+            f"DELETE FROM {self.table_name} WHERE id = ?",
+            (id_,)
+        )
+        self._connection.commit()
```
petrykivd, line 41: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM clause from new line.
In other places too

app/managers.py
```diff
@@ -0,0 +1,41 @@
+import sqlite3
+from typing import List
+
+
+from app.models import Actor
```
petrykivd, line 5: Fix the order of imports

app/managers.py
```diff
@@ -0,0 +1,41 @@
...
+    def all(self) -> List[Actor]:
+        cursor = self._connection.execute(
+            f"SELECT * FROM {self.table_name}"
+        )
+
+        return [Actor(*row) for row in cursor]
+
+    def update(self, id_: int, first_name: str, last_name: str) -> None:
+        self._connection.execute(
+            f"UPDATE {self.table_name} "
```
petrykivd, line 30: Maybe it's better to use actor_id than id_?

petrykivd (changes requested):

PR #1122, opened 2025-01-06
actors-solution

Bleit1:


10kkyvl (approved):

app/managers.py
```diff
@@ -0,0 +1,34 @@
...
+        with sqlite3.connect(self._DB_PATH) as conn:
+            actors_cursor = conn.execute(
+                f"SELECT * FROM {self._TABLE_NAME}")
+            return [Actor(*row) for row in actors_cursor]
+
+    def create(self, _first_name : str, _last_name: str) -> None:
+        with sqlite3.connect(self._DB_PATH) as conn:
+            conn.execute(
+                f"INSERT INTO {self._TABLE_NAME} VALUES (?, ?)",
+                (_first_name, _last_name))
```
petrykivd, line 21: Instead of creating a new connection in each etode, you can create one in the class initial and use it for .execute().
Much less code

app/managers.py
```diff
@@ -0,0 +1,34 @@
...
+class ActorManager:
+    _DB_PATH = "cinema.sqlite"
+    _TABLE_NAME = "actors"
+
+    def all(self) -> list[Actor]:
+        with sqlite3.connect(self._DB_PATH) as conn:
+            actors_cursor = conn.execute(
+                f"SELECT * FROM {self._TABLE_NAME}")
+            return [Actor(*row) for row in actors_cursor]
+
```
petrykivd, line 16: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM clause from new line.
Fix it in all places too

app/managers.py
```diff
@@ -0,0 +1,34 @@
...
+                (_first_name, _last_name))
+
+    def update(self, _id : int, _first_name : str, _last_name: str) -> None:
+        with sqlite3.connect(self._DB_PATH) as conn:
+            conn.execute(
+                f"UPDATE {self._TABLE_NAME} "
+                f"SET first_name = ?, last_name = ? WHERE id = ?",
+                (_first_name, _last_name, _id))
+
+    def delete(self, _id : int) -> None:
```
petrykivd, line 30: Maybe it's better to use actor_id instead of _id? 
And why do you need underscores for last_name and first_name? Are these protected parameters? 😮

petrykivd (changes requested):

app/managers.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class ActorManager:
+    def __init__(self, db_path: str, table_name: str) -> None:
+        self._db_path = db_path
+        self._table_name = table_name
+        self.conn = sqlite3.connect(self._db_path)
+
+    def all(self) -> list[Actor]:
+        cursor = self.conn.execute(f"SELECT * FROM {self._table_name}")
```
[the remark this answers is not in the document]
petrykivd: Fix the comments on SQL code style

petrykivd (changes requested):

L1nk3rrr (approved):
Well done!

PR #1130, opened 2025-01-07
Solution

Professor-Douel:


app/main.py
```diff
@@ -1,9 +1,9 @@
...
-#
+from models import Actor
+from managers import ActorManager
+
+if __name__ == "__main__":
+    Actor.objects = ActorManager()
+
 #     Actor.objects.create(first_name="Emma", last_name="Watson")
 #     Actor.objects.create(first_name="Daniel", last_name="Radclife")
 #     print(Actor.objects.all())
```
petrykivd, line 9: Should this be commented on?

app/managers.py
```diff
@@ -0,0 +1,48 @@
...
+        self._connection.execute(
+            f"INSERT INTO {self.table_name}"
+            f"(first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> list:
+        actor_cursor = self._connection.execute(
+            f"SELECT * FROM {self.table_name}"
```
petrykivd, line 25: It's better to follow SQL style guidelines even while writing raw SQL in Python. Therefore I'd suggest to start FROM clause from new line.

petrykivd: Fix it in all places

petrykivd (changes requested):

petrykivd (approved):
GJ! 😎

PR #1132, opened 2025-01-07
Solution

shretsker:


app/main.py
```diff
@@ -1,13 +1,13 @@
...
-#     Actor.objects.delete(1)
-#     print(Actor.objects.all())
+from models import Actor
+from managers import ActorManager
+
+if __name__ == "__main__":
+    Actor.objects = ActorManager()
+
+    Actor.objects.create(first_name="Emma", last_name="Watson")
+    Actor.objects.create(first_name="Daniel", last_name="Radcliffe")
```
Professor-Douel, line 8: You dont need to change value "Radclife" it is a condition of the task, you will change it in update code after.

app/managers.py
```diff
@@ -0,0 +1,54 @@
+import sqlite3
+
+# from models import Actor
```
Professor-Douel, line 3: Why do you comment the import? Your code working without import?

app/managers.py
```diff
@@ -0,0 +1,54 @@
...
+            f"INSERT INTO {self.table_name}"
+            f" (first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+# retrieve
+    def all(self) -> list:
+        actors_cursor = self._connection.execute(
+            "SELECT * FROM actors"
```
Professor-Douel, line 23: You have the init in your class: self.table_name and here "SELECT * FROM actors" you can use "SELECT * FROM {self.table_name}"

Professor-Douel (approved):

PR #1163, opened 2025-01-31
Created all classes and moduls

shafran-andriy:


app/managers.py
```diff
@@ -0,0 +1,57 @@
...
+    def initialize_db(self) -> None:
+        self._connection.execute(
+            f"CREATE TABLE IF NOT EXISTS {self.table_name}"
+            f" (id INTEGER PRIMARY KEY AUTOINCREMENT, "
+            f" first_name TEXT NOT NULL, "
+            f" last_name TEXT NOT NULL)"
+        )
+        self._connection.commit()
+
+    # CREATE - C
```
vasylhnatiuk, line 21: delete all comments, it is clear without them

vasylhnatiuk (changes requested):

app/managers.py
shafran-andriy: all comments were deleted

vsmutok (changes requested):

app/managers.py
shafran-andriy: removed and added the file cinema.sqlite to .gitignore

vsmutok (approved):
Looks Good!

PR #1195, opened 2025-02-21
Solution

kram3ko:


app/cinema.db
oliasymak24, line 1: You don't need to push your bd, delete it from pull request

oliasymak24 (changes requested):

PR #1218, opened 2025-03-11
Completed task py-actor-manager

Yurkiv313:


app/main.py
```diff
@@ -1,13 +1,13 @@
...
+if __name__ == "__main__":
+    Actor.objects = ActorManager()
+
+    Actor.objects.create(first_name="Emma", last_name="Watson")
+    Actor.objects.create(first_name="Daniel", last_name="Radclife")
+    print(Actor.objects.all())
+    Actor.objects.update(2, "Daniel", "Radcliffe")
+    print(Actor.objects.all())
+    Actor.objects.delete(1)
+    print(Actor.objects.all())
```
Baranotik15, line 13: When using the manager you do not close the connection, use the context manager or close it manually

app/managers.py
```diff
@@ -0,0 +1,47 @@
+import sqlite3
+
+from models import Actor
+
+
+class ActorManager:
+
+    def __init__(self) -> None:
+        self._connection = sqlite3.connect("cinema_db.sqlite")
+        self.table_name = "actors"
```
Baranotik15, line 10: Here you need to add a field to close the connection

app/managers.py
```diff
@@ -0,0 +1,47 @@
...
+            "WHERE id=? ",
+            (new_first_name, new_last_name, up_to_update)
+        )
+        self._connection.commit()
+
+    def delete(self, id_to_delete: int) -> None:
+        self._connection.execute(
+            f"DELETE FROM {self.table_name} WHERE id= ? ",
+            (id_to_delete,)
+        )
```
Baranotik15, line 47: How does the code work if you pass an id that does not exist in the database?

Baranotik15, line 47: Also after deletion you need to commit the changes to the database.
Use         
self._connection.commit()
like in Update in delete method

app/managers.py
```diff
@@ -0,0 +1,47 @@
+import sqlite3
+
+from models import Actor
```
Baranotik15, line 3: it is more correct to use absolute import

from app.models import Actor

Baranotik15 (changes requested):
Gj but need some changes.

app/main.py
Yurkiv313: There was nothing about closing the connection in this topic, look in your project to see if you used closing the connection, and this file was not created by me.

app/managers.py
Yurkiv313: agree

oliasymak24 (approved):
Good job!

Baranotik15 (approved):

PR #1267, opened 2025-04-02
'Solution'

Sezaro104:


app/managers.py
```diff
@@ -0,0 +1,47 @@
...
+           """)
+        self.conn.commit()
+
+    def create(self, first_name: str, last_name: str) -> int:
+        self.cursor.execute("INSERT INTO actors (first_name, last_name)"
+                            "VALUES (?, ?)", (first_name, last_name))
+        self.conn.commit()
+        return self.cursor.lastrowid
+
+    def all(self) -> list:
```
Boyok1337, line 27: For accuracy, you can add that these are list[Actor]

Boyok1337 (approved):

PR #1273, opened 2025-04-06
Created database manager by CRUD

Tolik-code:


app/managers.py
```diff
@@ -0,0 +1,37 @@
+import sqlite3
+
+from app.models import Actor
+
+
+class ActorManager:
+    def __init__(self) -> None:
```
Arch0998, line 7: The task required you to create in the init method, the class should accept two parameters: db_name (the name of the database) and table_name (the name of the table). The init method should also establish a connection to the database.

Arch0998 (changes requested):

PR #1279, opened 2025-04-11
Develop

olek007:


olek007:
@mateacademy-ai-mentor 
The database path is not hardcoded here. You can pass argument "db_path" to change db.
actors is plural

PR #1407, opened 2025-07-23
'RESULT'

DanilHamster:


app/managers.py
```diff
@@ -3,4 +3,46 @@
...
+        )
+        self._connection.commit()
+
+    def delete(self, pk: str) -> None:
+        self._connection.execute(
+            f"DELETE FROM {self.table_name} "
+            "WHERE id = ?",
+            (pk,)
+        )
+        self._connection.commit
```
DrGregoryHouse4676: ❌ this is a method reference, not a call, try adding parentheses.(self._connection.commit)

app/managers.py
```diff
@@ -3,4 +3,46 @@
...
+               last_name: str
+               ) -> None:
+        self._connection.execute(
+            f"INSERT INTO {self.table_name} "
+            f"(first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name)
+        )
+        self._connection.commit()
+
+    def all(self) -> None:
```
DrGregoryHouse4676: The all method has the annotation -> None, but returns a list, you can use the annotation list[Actor]

DrGregoryHouse4676 (approved):

PR #1414, opened 2025-07-24
Solution

farzamgt:


VadymKyrylko (changes requested):
Overall, good job, you did much more than the condition required, but it is better to write each function argument on a new line

PR #1415, opened 2025-07-24
Solution

VadymKyrylko:


Artcrafterrra (approved):
Good job, all done according to the task! Just one minor mistake with the annotation, which is not necessary in general.

PR #1417, opened 2025-07-24
Solution

Artcrafterrra:


Vskesha (approved):
Simple and elegant solution. Great work. Without any comments

Artcrafterrra:
@mateacademy-ai-mentor The SQL statement is syntactically correct—Python automatically concatenates adjacent string literals, even across lines. So the query works as expected. However, I can rewrite it as a single string if that’s preferred for clarity or to meet style guidelines.

PR #1425, opened 2025-07-25
Solution

dimashevchukk:


viannik (approved):

dimashevchukk:
@mateacademy-ai-mentor self._connection.commit() is already called after the delete

PR #1433, opened 2025-07-28
Solution

Irahromska:


app/managers.py
```diff
@@ -1,73 +1,44 @@
...
-            return [
-                Actor(
-                    id=row["id"],
-                    first_name=row["first_name"],
-                    last_name=row["last_name"]
-                ) for row in rows
-            ]
+    def update(
+        self,
+        pk: str,
```
trellle, line 27: primary key need to be integer

app/managers.py
```diff
@@ -1,73 +1,44 @@
...
-            cursor.execute(
-                f"DELETE FROM {self.table_name} WHERE id = ?", (pk,)
-            )
-            conn.commit()
+    def delete(self, pk: str) -> None:
+        self._connection.execute(
+            f"DELETE FROM {self.table_name} WHERE id = ?",
+            (pk,)
+        )
+        self._connection.commit()
```
trellle, line 44: it will be better if u close the connection in destructor

trellle (approved):

PR #1487, opened 2025-09-27
Solution

delirium95:


app/managers.py
```diff
@@ -3,4 +3,40 @@
 from app.models import Actor
 
 
-# add manager here
+class ActorManager:
+    def __init__(self, db_name: str, table_name: str) -> None:
+        self.db_name = db_name
+        self.table_name = table_name
+        self.connection = sqlite3.connect(self.db_name)
+        self.table_name = table_name
```
Vorongor, line 11: Why do you store table_name a second time?

app/managers.py
```diff
@@ -3,4 +3,40 @@
 from app.models import Actor
 
 
-# add manager here
+class ActorManager:
+    def __init__(self, db_name: str, table_name: str) -> None:
+        self.db_name = db_name
+        self.table_name = table_name
+        self.connection = sqlite3.connect(self.db_name)
```
Vorongor, line 10: Would it be better to make all params protected?

app/managers.py
```diff
@@ -3,4 +3,40 @@
...
+        )
+        self.connection.commit()
+
+    def all(self) -> list[Actor]:
+        cursor = self.connection.execute(
+            f"SELECT id, first_name, last_name FROM {self.table_name}"
+        )
+        rows = cursor.fetchall()
+        return [Actor(id=row[0], first_name=row[1],
+                      last_name=row[2]) for row in rows]
```
Vorongor, line 27: It a greate option use unpacking here insted of list elements

Vorongor (approved):
Well done, but there are some necessary mistakes that need to be corrected!

PR #1536, opened 2025-11-20
Actor-Manager

n-korobko:


VadymVynokur (approved):
You're using some methods I'm not familiar with, so I can't understand this solution, but it definitely works, so great job!

PR #1537, opened 2025-11-20
add solution

VadymVynokur:


app/managers.py
```diff
@@ -3,4 +3,38 @@
...
+
+    def __init__(self, db_name: str, table_name: str) -> None:
+        self._connection = sqlite3.connect(db_name)
+        self.table_name = table_name
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self._connection.execute(f"INSERT "
+                                 f"INTO {self.table_name} "
+                                 f"(first_name, last_name) "
+                                 f"VALUES ('{first_name}', '{last_name}')")
```
vadym-zahrebelnyi: You should also use parameterization here for first_name and last_name, and the same situation for deletion method

app/managers.py
```diff
@@ -3,4 +3,38 @@
...
+            self._connection.execute(
+                f"SELECT * FROM {self.table_name}")
+        )
+        return [Actor(*row) for row in actor_manger_cursor]
+
+    def update(self, pk: int, new_first_name: str, new_last_name: str) -> None:
+        self._connection.execute(
+            f"UPDATE {self.table_name} "
+            f"SET first_name = ? "
+            f"WHERE id = ?", (new_first_name, pk))
```
vadym-zahrebelnyi: f-string is needed only for the table_name; you have not to do that for continuous segments of the query with parameterization

vadym-zahrebelnyi (approved):
Well done!

PR #1553, opened 2025-11-25
actor-manager

ohorodnykostap:
actor-manager


app/managers.py
```diff
@@ -1,6 +1,38 @@
...
+from typing import List
 from app.models import Actor
 
 
-# add manager here
+class ActorManager:
+    def __init__(self, db_name: str, table_name: str) -> None:
+        self.db_name = db_name
+        self.table_name = table_name
+        self._connection = sqlite3.connect(self.db_name)
```
tradewcs, line 10: You are not closing connection anywhere

tradewcs (changes requested):

app/managers.py
ohorodnykostap: Done.

tradewcs (approved):

PR #1738, opened 2026-04-20
Implemented class Actor in models.py, class ActorManager in managers.py

artradkevych:
Implemented class Actor in models.py, class ActorManager in managers.py.
Added CRUD methods(create, all, update, delete) to ActorManager


AnyoneClown (approved):
LGTM

PR #1742, opened 2026-04-25
Implemented ActorManager

PavelPereverzev1:


vova-dev-python (approved):
Great job on the CRUD logic and the use of parameterized queries! To make the code even more robust, consider naming the columns explicitly in the all method instead of using *.

PR #1761, opened 2026-04-29
Solution

smykroman-dot:


app/managers.py
```diff
@@ -1,6 +1,65 @@
...
+    def all(self) -> List[Actor]:
+        self.cursor.execute(
+            f"SELECT id, first_name, last_name FROM {self.table_name}"
+        )
+        rows = self.cursor.fetchall()
+
+        return [
+            Actor(id=row[0], first_name=row[1], last_name=row[2])
+            for row in rows
+        ]
```
petrykivd, line 42: You can use unpacking insted of using indexes

app/managers.py
```diff
@@ -1,6 +1,65 @@
...
+        self.cursor.execute(
+            f"INSERT INTO {self.table_name} "
+            f"(first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name),
+        )
+        self.connection.commit()
+
+    def all(self) -> List[Actor]:
+        self.cursor.execute(
+            f"SELECT id, first_name, last_name FROM {self.table_name}"
```
petrykivd, line 35: Let's use everywhere SQL style (every operation from new string)

app/managers.py
```diff
@@ -1,6 +1,65 @@
...
+
+    def create(self, first_name: str, last_name: str) -> None:
+        self.cursor.execute(
+            f"INSERT INTO {self.table_name} "
+            f"(first_name, last_name) VALUES (?, ?)",
+            (first_name, last_name),
+        )
+        self.connection.commit()
+
+    def all(self) -> List[Actor]:
```
petrykivd, line 33: Using list instead of List is preferable in new python versions

petrykivd (changes requested):

